PR #11, opened 2016-09-17
New, more uniform stars mobject

Helpsypoo:
The old stars mobject used polar coords to generate points, so the points were concentrated around the origin. This one uses cartesian coords to make it more uniform. It also inherits from Mobject2D, since I tried that and it made it work, even though I don't know why.

Pretty trivial, but hey, I gotta fill up my green square grid!

### DIFF
--- a/topics/geometry.py
+++ b/topics/geometry.py
@@ -85,7 +85,9 @@ class Line(VMobject):
         start_to_end = self.end - self.start
         length = np.linalg.norm(start_to_end)
         if length > 2*self.buff:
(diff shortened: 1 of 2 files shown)


Helpsypoo:
Also recreated the octohedron class to inherit from VMobject.

I'm realizing now that I could have managed branches differently and created different pull requests. I can rejigger things if you'd like to accept one change but not the other.

3b1b:
3D things work a bit differently now, and will likely change again soon, so I'm closing this for now.

PR #21, opened 2017-04-17
Merge nested configs instead of overriding

styxtwo:
Changing background color by adding config to a scene is not possible. it is overridden by the config defined in constants (LOW_QUALITY_CAMERA_CONFIG), even when the constant does not define the color. This happens because the "camera_config" is nested, so its content is never compared.

With this change nested configuration is merged instead of overridden. The change is tested on several scenes from hanoi.py, and does not break any of the tested scenes.

### DIFF
--- a/helpers.py
+++ b/helpers.py
@@ -229,7 +229,6 @@ def filtered_locals(local_args):
         result.pop(arg, local_args)
     return result
 
-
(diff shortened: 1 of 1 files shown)


3b1b:
Thanks!

PR #26, opened 2017-05-04
Handle latex errors when converting to dvi

hashar:
Whenever latex fails (for example due to a missing package) the code
continue and ends up failing due to a missing svg file.

Handle latex non 0 exit code.
Pass to latex -halt-on-error so it does not attempt to ask user for a
file, that is ignored in batch mode anyway.

Write the latex output to stderr and raise an exception to abort early.

Signed-off-by: Antoine Musso <hashar@free.fr>

### DIFF
--- a/mobject/tex_mobject.py
+++ b/mobject/tex_mobject.py
@@ -3,6 +3,7 @@ from svg_mobject import SVGMobject, VMobjectFromSVGPathstring
 from topics.geometry import BackgroundRectangle
 from helpers import *
 import collections
+import sys
 
 TEX_MOB_SCALE_FACTOR = 0.05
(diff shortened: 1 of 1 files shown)


bhamshu:
I *am* getting the error as per the change you made on line 283 in 0ec5538ef0f0cb35a1a2a4e720d4d473901e6950, i.e. "Exception: Latex error converting to dvi." and so forth. How to resolve it?

hashar:
@bhamshu I can't remember the error I had two months ago. The reason for this patch is solely to abort and display the error immediately instead of having to crawl from the error log files :-)

Try to fill an issue?  Maybe you are missing some texlive packages.

PR #62, opened 2017-12-22
Ability to export PNG sequence

mitkonikov:
While writing to a movie file, if you specify -f it will export a png
sequence in a folder with the same name like the scene. The ability to
export transparent PNG sequences adds the ability to change the
background afterwards.

### DIFF
--- a/extract_scene.py
+++ b/extract_scene.py
@@ -24,6 +24,8 @@ HELP_MESSAGE = """
    -m use medium quality
    -a run and save every scene in the script, or all args for the given scene
    -q don't print progress
+   -f when writing to a movie file, export the frames in png sequence
+   -t use transperency when exporting images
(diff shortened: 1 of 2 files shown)


mitkonikov:
You may ask what will be the purpose of this feature, well..
#1: I made a video where I used some of the animations with the background being clips that I recorded with my DSLR: https://www.youtube.com/watch?v=jR3haKZtA5E
#2: Rendering time with background image in the Manim is sooo slow... Now, if you render the background transparent and afterwards in Premiere Pro add the desired background image, the combined rendering time will be much faster.

3b1b:
By the way, I justed change the relevant flag here from -f to -g, since I wanted -f to be used for showing the file in finder upon writing a file.

PR #76, opened 2018-01-19
Winding number

Sridhar3b1b:
Having made fixes to the implementation of Succession animations, these fixes should be pulled into master.


3b1b:
Looks good to me!

PR #83, opened 2018-01-22
A few improvements on TexMobject

Solara570:
1. Modified `highlight_by_tex_to_color_map()` so that it's able to accept a dictionary with both string and tuple as keys.
For example, the following color_map is now valid:
`{"x": BLUE, ("y", "z"): YELLOW}`
2. Also, the keyword argument `substring` should work properly in `highlight_by_tex_to_color_map()`.
3. Added some special characters in `modify_special_strings()`.


3b1b:
Thanks so much!

PR #85, opened 2018-01-23
New methods for Arcs and Sectors

bhbr:


3b1b:
Just one quick requested change before I merge this: For get_arc_center, implement it as self.point_from_proportion(0.5), just to keep things conceptually simpler.

bhbr:
Hang on - doesn't that just give the point halfway along the arc? The arc center is supposed to be the point around which the arc is turning (center of the prolonged circle).

bhbr:
How about we call the other point get_arc_midpoint or get_midpoint?

PR #107, opened 2018-02-01
ProjectAlongVector

bhbr:
A Transform that projects a mobject onto the plane with given normal vector (and passing through the origin).


3b1b:
Looks good!  Can you just exclude the ThreeDExample scene file from this pull request.  Or better yet, put it inside the example_scenes.py folder.

PR #109, opened 2018-02-02
[3D CAMERA] Center of rotation

frozar:
New feature: give the possibility to change the center of rotation.


frozar:
I add the example `Rotation3d` so you can check it:
```sh
python extract_scene.py -l example_scenes.py Rotation3d
```

Issue #143, opened 2018-03-03
Is there any way to animate Spheres or other curved 3D objects?


ddxtanx:
I've tried to use the Sphere() class in topics/three_dimensions.py but, by default, it is very 'point'-y and not shaded. How could I generate something similar to the spheres and curved 3D objects in 3B1B videos?

domino566:
There's an old_projects/three_dimensions.py file with a class Sphere and a function generate_points, could this be what you're looking for?
```
class Sphere(Mobject2D):
    def generate_points(self):
        self.add_points([
            (
                np.sin(phi) * np.cos(theta),
                np.sin(phi) * np.sin(theta),
                np.cos(phi)
            )
            for phi in np.arange(self.epsilon, np.pi, self.epsilon)
            for theta in np.arange(0, 2 * np.pi, 2 * self.epsilon / np.sin(phi)) 
        ])
        self.set_color(BLUE)

    def unit_normal(self, coords):
        return np.array(coords) / np.linalg.norm(coords)
```

3b1b:
At the moment, there is not.  I have been using Grapher to generate most curved 3d visuals.  Pull requests accepted though :)

PR #145, opened 2018-03-06
Add 3D graphing via square approximation.

ddxtanx:
While this isn't perfect, it is a way to graph a 3D surface using square approximations. (Also apparently some linting got in, which I can change if necessary.)

### DIFF
--- a/topics/three_dimensions.py
+++ b/topics/three_dimensions.py
@@ -1,4 +1,6 @@
+
 from helpers import *
+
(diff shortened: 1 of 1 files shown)


ddxtanx:
Meant to address #143

3b1b:
Really what's needed here is a different type of mobject built on a 3d graphics library.  Eventually, it'll be worth taking the time to implement that, but trying to make the vectorized Mobjects behave like curved surfaces is just going to be way too slow, and likely not great looking.

PR #241, opened 2018-05-24
Added a way to easily customize the media directory

rajdakin:
This pull request adds an easy way to reconfigure the media directories:
- You can have the environment variable MEDIA_DIR set, in which case it will use this and remember it
- You can have already set once the MEDIA_DIR environment variable, in which case it will use the last set
- Otherwise, it will defaults to a default directory.
In all cases, it will remember the choice and write it in a file (which is used in case you don't have the MEDIA_DIR variable set).

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -1,12 +1,40 @@
 import os
 import numpy as np
 
-# Things anyone wishing to use this repository for their
-# own use will want to change
-MEDIA_DIR = os.path.join(
-    os.path.expanduser('~'),
-    "Dropbox (3Blue1Brown)/3Blue1Brown Team Folder"
-)
+env_MEDIA_DIR = None
+MEDIA_DIR = "#ERROR#"
+
+try:
+    env_MEDIA_DIR = os.getenv("MEDIA_DIR")
+except NameError:
+    try:
+        env_MEDIA_DIR = os.environ['MEDIA_DIR']
+    except KeyError:
+        pass
+
+if not (env_MEDIA_DIR is None):
+    MEDIA_DIR = env_MEDIA_DIR
+elif os.path.exists("media_dir.txt"):
+    with open("media_dir.txt", 'rU') as media_file:
+        MEDIA_DIR = media_file.readline()
+    if MEDIA_DIR[-1] in ['\n', '\r']:
+        raise Exception("media_dir.txt must be a single-line text file pointing at the media_dir place")
+else:
+    MEDIA_DIR = os.path.join(
+        os.path.expanduser('~'),
+        "Dropbox (3Blue1Brown)/3Blue1Brown Team Folder"
+    )
+
+if not os.path.exists(MEDIA_DIR):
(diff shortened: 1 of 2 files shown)


constants.py
```diff
@@ -1,12 +1,40 @@
...
+    except KeyError:
+        pass
+
+if not (env_MEDIA_DIR is None):
+    MEDIA_DIR = env_MEDIA_DIR
+elif os.path.exists("media_dir.txt"):
+    with open("media_dir.txt", 'rU') as media_file:
+        MEDIA_DIR = media_file.readline()
+    if MEDIA_DIR[-1] in ['\n', '\r']:
+        raise Exception("media_dir.txt must be a single-line text file pointing at the media_dir place")
```
not7cd: why not
```python
with open("media_dir.txt", 'rU') as media_file:
    MEDIA_DIR = media_file.readline().strip()
```

not7cd: Tools like `pipenv` or `docker-compose` load env variables from `.env` files. I find this file loading complicating this already long file.
With adoption of `.env` you could expect that user can handle variables.

constants.py
```diff
@@ -1,12 +1,40 @@
...
+env_MEDIA_DIR = None
+MEDIA_DIR = "#ERROR#"
+
+try:
+    env_MEDIA_DIR = os.getenv("MEDIA_DIR")
+except NameError:
+    try:
+        env_MEDIA_DIR = os.environ['MEDIA_DIR']
+    except KeyError:
+        pass
```
not7cd: There are 2 ways you try getting env. Is this legitimate way to get it?

constants.py
```diff
@@ -1,12 +1,40 @@
...
+
+if not (env_MEDIA_DIR is None):
+    MEDIA_DIR = env_MEDIA_DIR
```
not7cd: Since environ is dictionary you could do this without helper variable
```python
if 'MEDIA_DIR' in os.environ:
     MEDIA_DIR = os.environ['MEDIA_DIR']
```
but I'm unsure about safety.

constants.py
rajdakin: 1. I just see with your comment that I wrote something stupid... The MEDIA_DIR cannot end with a new line. Anyways, I'll use that `strip()` command too, because it seems to be useful.
2. Yes, but manim must accept to add this dependancy, and I don't know how to use `pipenv` nor `docker-compose`.

constants.py
rajdakin: I know that I could just use the `os.environ` dictionary, but I prefer using the helpers methods, and even if `os.environ` does not exists I'd like the code not to crash... (see my previous answer)

constants.py
rajdakin: I think it's best to allow multiple possibilities to get an environment variable. (Note: `os.getenv(...)` is Windows/Unix only and `os.environ[...]` *may* be unstable.)

PR #242, opened 2018-06-02
Modernize Python 2 code to get ready for Python 3 yet again

cclauss:
Yet another shot at #183 and #189 

Some minimal, "safe" changes to make the code syntax compatible with both Python 2 and Python 3.
* Run __futurize --stage1 -w .__  # http://python-future.org/futurize_cheatsheet.html

### DIFF
--- a/old_projects/hilbert/section1.py
+++ b/old_projects/hilbert/section1.py
@@ -101,7 +101,7 @@ class AboutSpaceFillingCurves(TransformOverIncreasingOrders):
         self.wait()
         self.play(*[
             ShimmerIn(mob)
-            for mob in infinity, rightarrow, N
+            for mob in (infinity, rightarrow, N)
         ] + [
             ApplyMethod(question_mark.next_to, rightarrow, UP),
         ])
@@ -540,11 +540,11 @@ class ThinkInTermsOfReverseMapping(Scene):
         arrow2 = Arrow(4*RIGHT+UP, dot2, color = color2, buff = 0.1)
         dot3, arrow3 = [
             mob.copy().shift(5*LEFT+UP)
-            for mob in dot1, arrow1
+            for mob in (dot1, arrow1)
         ]
         dot4, arrow4 = [
             mob.copy().shift(5*LEFT+0.9*UP)
-            for mob in dot2, arrow2
+            for mob in (dot2, arrow2)
         ]
 
         self.add(grid, freq_line, arrow)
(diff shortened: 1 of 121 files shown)


old_projects/hilbert/section1.py
```diff
@@ -659,7 +659,7 @@ def construct(self):
         mathy, bubble = get_mathy_and_bubble()
         squiggle_mouth = mathy.mouth.copy()
         squiggle_mouth.apply_function(
-            lambda (x, y, z) : (x, y+0.02*np.sin(50*x), z)
+            lambda x_y_z : (x_y_z[0], x_y_z[1]+0.02*np.sin(50*x_y_z[0]), x_y_z[2])
```
not7cd: Is this more pythonic?

old_projects/matrix_as_transform_2d.py
```diff
@@ -93,9 +93,11 @@ def construct(self):
 
 class ExamplesOfNonlinearOneDimensionalTransforms(NumberLineScene):
     def construct(self):
-        def sinx_plux_x((x, y, z)):
+        def sinx_plux_x(xxx_todo_changeme):
+            (x, y, z) = xxx_todo_changeme
```
not7cd: maybe `xxx_todo_changeme` as `vector3`?

old_projects/hilbert/section1.py
cclauss: It is not about being Pythonic.  It is about avoiding syntax errors.  Putting the parameters to a lambda in parens is a syntax error in Python 3.

zsteinberg: Multiple arguments to a lambda isn't a syntax error if you just remove the parentheses:
    >>> func = lambda x,y,z: x+x
    >>> func(1,2,3)
    2

not7cd: But then there are multiple places in code that expect that this lambdas need tuples as parameters. It could be change later if only thing is better python 3 compatibility without changing workflow. @3b1b?

cclauss: There are two ways that the caller can call the lambda:
1. With one single tuple: __call(x_y_z)__
2. With one multiple parameters: __call(x, y, z)__

You either need to make sure that you know how all callers (both inside and outside the codebase) will call the lambda --or-- you need to gracefully support _both_ the calling conventions.

PR #283, opened 2018-09-06
Fixed a typo

nerkulec:
Fixed a typo in line 27 (TexMobject -> TextMobject)


3b1b:
Actually, that's not a typo.  The TexMobject type produces LaTeX expressions, like equations, where as TextMobject produces similar objects meant to be outside math mode.

PR #301, opened 2018-09-18
update Dockerfile for python 3

eulertour:
The current Dockerfile was written for python 2, which is no longer supported. This updates it for python 3.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,37 @@
-FROM python:2.7.12
-RUN echo "deb http://ftp.debian.org/debian jessie-backports main" >> /etc/apt/sources.list
-RUN apt-get update && apt-get install -y \
-    git
+FROM ubuntu:18.04
+ENV DEBIAN_FRONTEND noninteractive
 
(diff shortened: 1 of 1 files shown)


cclauss:
The previous version did __FROM python:2.7.12__ so why not start this version with __FROM python:3.7.0__ as that would seem to save a lot of __apt get install__ time and complexity. That team can manage Python packaging so we don’t have to.  Also would it be possible to consolidate all of the separate calls to __apt get install__ into just one or two calls?  It is currently a lot for the reader to digest.

eulertour:
> The previous version did **FROM python:2.7.12** so why not start this version with **FROM python:3.7.0** as that would seem to save a lot of **apt get install** time and complexity? That team can manage Python packaging so we don’t have to. Also would it be possible to consolidate all of the separate calls to **apt get install** into just one or two calls? It is currently a lot for the reader to digest.

Sorry I didn't see this before. I'd intended for the dockerfile to be a template for how one would install each of the individual dependencies step by step. This might be worth changing when the dependencies are clearly laid out in the README or somewhere else.

PR #303, opened 2018-09-19
updated some breaking scenes to python3

ishandutta2007:
These changes would fix few(100+) scenes for python 3.x in both old_projects as well as active_projects directory


icreiuheciijc:
Once I cloned manim after this every TextMobject("Something") would create tex files "S", "o", "m", "e", ... , "g", rather than "Something", and it is due to change to file `mobject/svg/tex_mobject.py`. Is it my problem that I should use TextMobject(["Something"]) instead now?

3b1b:
That line turned out not to be necessary (a vestige of an older time I think), and removing it fixes the error.

Issue #308, opened 2018-09-21
SyntaxError: invalid syntax in return np.array([*color_to_rgb(color), alpha])


CRLannister:
while running python extract_scene.py example_scenes.py SquareToCircle -pl in a vitualenv with python 2.7.15
The error output is: 
Traceback (most recent call last):
  File "extract_scene.py", line 15, in <module>
    from scene.scene import Scene
  File "/home/ashish/Documents/Codes/python/3b1b-manim/manim/scene/scene.py", line 15, in <module>
    from animation.animation import Animation
  File "/home/ashish/Documents/Codes/python/3b1b-manim/manim/animation/animation.py", line 5, in <module>
    from mobject.mobject import Mobject
  File "/home/ashish/Documents/Codes/python/3b1b-manim/manim/mobject/mobject.py", line 14, in <module>
    from utils.color import color_gradient
  File "/home/ashish/Documents/Codes/python/3b1b-manim/manim/utils/color.py", line 23
    return np.array([*color_to_rgb(color), alpha])
                     ^
SyntaxError: invalid syntax

cclauss:
[flake8](http://flake8.pycqa.org) testing of https://github.com/3b1b/manim on Python 3.7.0

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./old_projects/bell.py:222:38: E999 SyntaxError: invalid syntax
            for here, x in (here1, 0), (here2, 4)
                                     ^
./old_projects/borsuk.py:217:11: E999 SyntaxError: invalid syntax
         y=lambda m: m1.get
          ^
./old_projects/eoc/chapter6.py:399:35: E999 SyntaxError: invalid syntax
                for i, j in (1, 0), (0, 1)
                                  ^
./old_projects/nn/network.py:258:16: F821 undefined name 'get_norm'
        norm = get_norm(gradient)
               ^
./old_projects/nn/network.py:264:15: F821 undefined name 'get_norm'
        print(get_norm(old_pre_sig_guess - pre_sig_guess))
              ^
3     E999 SyntaxError: invalid syntax
2     F821 undefined name 'get_norm'
5
```

Elteoremadebeethoven:
manim-master works in Python 3, not in Python 2, install pip3 and requieriments with pip3

cclauss:
I believe that this issue can be closed.

Matemasters:
Greetings!
I'm already in Python 3.6.6 and when I type 

[...]

PR #309, opened 2018-09-22
README.md: Ensure that Python 3 is used

cclauss:
To avoid issues like #308

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ operating system:
 
 Then you can install the python dependencies:
 ```sh
-pip install -r requirements.txt
+python3 -m pip install -r requirements.txt
 ```
 
 ## How to Use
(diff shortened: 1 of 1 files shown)


icreiuheciijc:
Or maybe tell people that Todd Zimmerman's tutorial is outdated? Because it seems a lot of people who use python2 are referring to his blog..

Alternatively maybe raise exception if sys.version starts with '2'

zimmermant:
I'll add a note to each blog post about the difference between the two versions.  I'd like to update the posts to the latest version but I don't know when I'll have time.

cclauss:
My sense is that this is an __and__ thing, not an __or__ thing.  Both the README __and__ the blogs should be updated so that both sources of information are in sync on the point that _legacy Python is passe_ for users of manim.

PR #318, opened 2018-10-01
Modify Directions for Docker to Reflect Interactivity

DanWaxman:
PR #301 changed the role of Docker in this project to an interactive shell, which slightly changes instructions. Additionally, since it's interactive I added vim to its build to allow for some lightweight editing inside the environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,7 @@ RUN apt-get install -qqy texlive-full
 RUN apt-get install -qqy texlive-fonts-extra
 RUN apt-get install -qqy sox
 RUN apt-get install -qqy git
+RUN apt-get install -qqy vim
 
 ENV DEBIAN_FRONTEND teletype
 ENTRYPOINT ["/bin/bash"]
--- a/README.md
+++ b/README.md
@@ -60,4 +60,4 @@ a Dockerfile provided.
 
 1. [Install Docker](https://www.docker.com/products/overview)
 2. Build docker image. `docker build -t manim .`
-3. Run it! `docker run --rm -v "$PWD/files":/app/files manim example_scenes.py WarpSquare`
+3. Run it! `docker run --rm -itv "$PWD/files":/app/files manim`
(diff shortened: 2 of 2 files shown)


eulertour:
I think it makes more sense to add vim to a personal docker image built from the one here rather than add it to the base image. There are a few reasons for this:

First, the dockerfile is currently an easy way to see all of manim's dependencies, which are not clearly listed in the README. Ideally it would contain everything manim needs and nothing it doesn't, in order to eliminate ambiguity as to what is required and what isn't.

Second, the choice to add vim is a rather personal one, and it would alienate people who prefer emacs, nano, etc. Setting this precedent of adding packages for convenience could make the image prohibitively large (and it's already pretty big due to TexLive).

Third, and I realize I may be getting ahead of myself with this one, I think we should be working toward making manim behave more like a library than a bundle of scripts. Including an editor in a library is bad for modularity at best, and an admission of poor code quality at worst.

I want to be clear that I agree with the other changes, and I think adding vim makes sense for a personal image. I just don't think it should be done here. That said, the final decision isn't up to me.

DanWaxman:
I largely agree with you. However, the reason I decided to add vim into the base image instead of keeping it in just my personal image is due to the stated purpose of Docker in this project right now; it seems that Docker here is meant mostly to set up dependencies that are otherwise difficult or unclear. In other words, it's meant to help people set things up that otherwise do not have advanced knowledge of Linux, etc. I think this is evidenced by a lot of the issues with Docker posted in this repo -- a lot of them are fairly simple, but posted by people who are just learning.

As a result, it might not be a great idea to leave out a text editor, as one of the target audiences of the Docker file might expect a more ready-to-go development environment. Now I admit that for the audience of beginners, vim might not be the best editor to include, but I think at least something like nano should be included. 

An alternative for this could be removing the install from the dockerfile and creating a setup script in manim/files or something to setup some popular development tools, but I'm not sure this is a great solution either.

Let me know what you think about this, because I do agree with your points, but also see a reason to include it.

eulertour:
I think the best solution is to update the README or a page linked from it with clearer information on how to set up and use manim. As it is, nothing about manim is ready-to-go, as is evidenced by the steady influx of issues due to neglecting to configure MEDIA_DIR, install dependencies, or use the correct version of Python. I just don't think we should address the problem by changing the one part of manim that offers any sort of guarantee about it.

PR #327, opened 2018-10-14
Fixed some old projects

rorik:
All neural network scenes should be working now on Python 3.7 (I tested about 80% of them).

Also fixed a few lines in some other projects.

A few things that need to be checked:

- [Changes to mobject.svg.tex_mobject.get_parts_by_tex method](https://github.com/rorik/manim/commit/802de194cbee1926c26434d3b61d4729645d09df#diff-56ab4ca0436da9da7ce94d8557ab647aR191), make sure this doesn't break anything else.
- [This line which might not do the same](https://github.com/rorik/manim/commit/802de194cbee1926c26434d3b61d4729645d09df#diff-fed2a389a6547800df59c13b8126c971R564).
- [Not sure if about_point is correct here](https://github.com/rorik/manim/commit/802de194cbee1926c26434d3b61d4729645d09df#diff-14e11afed910dadaad65fff00142faa0R2947).

### DIFF
--- a/mobject/svg/tex_mobject.py
+++ b/mobject/svg/tex_mobject.py
@@ -189,7 +189,10 @@ class TexMobject(SingleStringTexMobject):
         return self
 
     def get_parts_by_tex(self, tex, substring=True, case_sensitive=True):
-        def test(tex1, tex2):
+        def test(tex1, object2):
+            if not isinstance(object2, SingleStringTexMobject):
+                return tex1 is None
+            tex2 = object2.get_tex_string()
             if not case_sensitive:
                 tex1 = tex1.lower()
                 tex2 = tex2.lower()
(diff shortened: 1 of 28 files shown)


rorik:
There's some visible differences between the output and the one that was generated using Python2.7 and uploaded to youtube.
## Expected:


## Actual output:


I'm going to look into what causes these differences and try to fix them. Any suggestions are welcome.

rorik:
It's mostly fixed.
## Actual output:


Albeit now the number neuron value at the start displays two decimal places instead of one, it could be fixed, but it's such a small detail that I'm not going to bother.
## Expected:


## Actual output:

Kolloom:
changes to the codebase itself will require 3b1b's approval. 

Also the old_projects and active_projects is not under a copyleft license. He recommends using old version of the project that correctly reproduce the ouput. 

I am not sure if there is a desire to update the old project to use the new code.

TonyCrane:
The projects of 3b1b have already been removed from this repository. So closed this.

PR #328, opened 2018-10-14
Cleanup Dockerfile

xfbs:
Thank you for making this tool available to everyone. This is just a tiny thing, the `Dockerfile` can be simplified by removing two packages that are installed anyways. The apt package texlive-full already includes texlive-latex-base and texlive-fonts-extra, so they don't have to installed separately.


eulertour:
If you want you can remove this line as well, since the entire Python3 directory is deleted later.
https://github.com/3b1b/manim/blob/8d66cfb75b2781e74f1de29d8e331f349aef03a8/Dockerfile#L21

xfbs:
@eulertour Good idea. I've also consolidated the `apt install` calls into one single one and sprinkled some comments into there. How does it look?

eulertour:
I think it's best not to consolidate the `apt install` commands because it obscures the logical flow of the installation (see [here](https://github.com/3b1b/manim/pull/301#issuecomment-425510200)). For example, the dockerfile won't work after 4ba9f6b5a31188c53cf51da618de351863a43348 is applied because `texlive` is installed before the time zone is setup (you'll get [this problem](https://askubuntu.com/questions/909277/avoiding-user-interaction-with-tzdata-when-installing-certbot-in-a-docker-contai)).

xfbs:
Welp, how's this?

eulertour:
The way it is now, `requirements.txt` will be left on the image since it gets copied in after the python 3 directory is deleted. Additionally, the time zone setup is still separated from the texlive install, which makes it unclear why it is necessary (although it was admittedly pretty unclear before as well).

I'd recommend moving the time zone setup to just before the texlive install, mentioning in the comment that the time zone setup is required by texlive, and getting rid of `requirements.txt` somehow, either by deleting it with the python 3 directory or adding another `rm` for it afterward.

xfbs:
Can you confirm that it now works as intended and that it makes sense?

eulertour:
This looks fine. The comment says textlive instead of texlive, but that's a nitpick.

Asday:
It's best to have all the `apt install`ing that van be done at once all be done in one `RUN` command.  That way it only creates one layer, saving the end user a bunch of disk space.

To make the list of dependencies clear, use backslashes:

```Dockerfile
RUN apt install -y \
  build-essential \
  something-else \
  yet-another-package
```

eulertour:
@Asday You're completely right, this dockerfile is ridiculous.

eulertour:
The good news is that the dockerfile is now much simpler. The bad news is that the image is actually _larger_ now, up to 5.72 GB from 4.28 GB.

PR #349, opened 2018-11-24
SVG single-factor scaling

sswatson:
Some fonts (like mathpazo) produce svg files which use single scale-factor transform (like `transform=scale(1.3)`). Previously the behavior of this code was to throw an unpacking exception in the try block, which resulted in the font glyphs misrendering. 

I would think that the `about_point` should be `ORIGIN` for the [scale_x, scale_y, 1] block as well, but I only put that change in the `elif` clause so that the proposed change would be non-breaking.

### DIFF
--- a/mobject/svg/svg_mobject.py
+++ b/mobject/svg/svg_mobject.py
@@ -264,8 +264,13 @@ class SVGMobject(VMobject):
             if not transform.startswith(prefix) or not transform.endswith(suffix):
                 raise Exception()
             transform = transform[len(prefix):-len(suffix)]
-            scale_x, scale_y = string_to_numbers(transform)
-            mobject.scale(np.array([scale_x, scale_y, 1]))
+            scale_values = string_to_numbers(transform)
+            if len(scale_values) == 2:
+                scale_x, scale_y = scale_values
+                mobject.scale(np.array([scale_x, scale_y, 1]))
+            elif len(scale_values) == 1:
+                scale = scale_values[0]
+                mobject.scale(np.array([scale,scale,1]), about_point = ORIGIN)
         except:
             pass
 


eulertour:
Does it make any difference if you leave out the optional argument to `mobject.scale()` and have `mobject.scale(np.array([scale,scale,1]))`? I ask because [`Mobject.scale()` scales about the origin by default anyway](https://github.com/3b1b/manim/blob/ec5c2acd3f72c9dd3f1139945bdf6826d9a38f86/mobject/mobject.py#L206) and I want to be sure I understand what this is doing.

sswatson:
It doesn't seem to work. I removed `about_point = ORIGIN`, and the result looks like this: 

screen shot 2018-12-04 at 11 24 49 am

Given that `ORIGIN` is supposed to be the default, I don't understand why that doesn't work. Here's a MWE (with `\usepackage{mathpazo}` added to `tex_template.tex`).  

```python
class Summation(Scene):
    def construct(self):
        T = TexMobject("\sum_{k=0}^\\infty \\frac{c}{k^2} =  \\frac{8\\pi^2}{996}")
        self.play(GrowFromCenter(T))
```

Perhaps the issue is a discrepancy between `ORIGIN`s? From the image, you can see that it looks like the summation symbol has been scaled from the bottom left.

sswatson:
Looked into this a bit more carefully: the center of the Mobject is not generally the same as ORIGIN. The SVG scaling does appear to be with respect to the origin, so `about_point = ORIGIN` is the desired behavior. I was able to confirm that this should be the case as well for the two-factor scaling (in other words, the fact that of `about_point = ORIGIN` is missing is a bug, in the sense that it causes the figures to render differently from how SVGs are rendered by the browser).

eulertour:
You're absolutely right.

PR #351, opened 2018-11-28
Prefix the link to the documentation with https:// .

H-M-H:
Otherwise GitHub sets the wrong prefix which results in:
https://github.com/3b1b/manim/blob/master/manim.readthedocs.io


eulertour:
Good catch.

PR #352, opened 2018-11-28
Fixed documentation link

AtHeartEngineer:
The previous link appended `manim.readthedocs.io` to the github url.


eulertour:
The linked page actually goes in the parentheses rather than the square brackets, so this doesn't fix the link.

Issue #401, opened 2019-01-13
The video doesn't concatenate, I can only get partial videos


yerovi84:
I have only the partial videos with the next error:

"[concat @ 000001ff22102900] Impossible to open '0.mp4'
media\videos\example_scenes\480p15\partial_movie_files\WriteStuff\partial_movie_file_list.txt: No such file or directory
File ready at media\videos\example_scenes\480p15\WriteStuff.mp4"

But I don't have the video WriteStuff.mp4.

Please help me

pdcxs:
Same issue. I use windows platform and get the same problem.

DRBrodbeck:
I also get this error.  What is the mystery text "partial_movie_file_list.txt" (from scene.py) below?

        file_list = os.path.join(
            partial_movie_file_directory,
            "partial_movie_file_list.txt"
        )

Elteoremadebeethoven:
The problem was fixed today in the original repository of 3b1b.

PR #402, opened 2019-01-13
The file problem on Windows

pdcxs:
Issue: https://github.com/3b1b/manim/issues/401
I use windows platform and when testing the example, ffmpeg cannot find 0.mp4 and the partial_movie_file_list.txt. Moreover, the partial files and directory are not deleted when finished.

Testing:
Before:
Command: python -m manim example_scenes.py SquareToCircle
tim 20190113234228

After: 
Same Command
tim 20190113234039

### DIFF
--- a/manimlib/scene/scene.py
+++ b/manimlib/scene/scene.py
@@ -720,7 +720,7 @@ class Scene(Container):
         with open(file_list, 'w') as fp:
             for pf_path in partial_movie_files:
                 if os.name == 'nt':
-                    fp_path = fp_path.replace('\\', '/')
+                    pf_path = pf_path.replace('\\', '/')
 
                 fp.write("file {}\n".format(pf_path))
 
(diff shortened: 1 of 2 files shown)


manimlib/scene/scene.py
```diff
@@ -719,6 +719,9 @@ def combine_movie_files(self):
         )
         with open(file_list, 'w') as fp:
             for pf_path in partial_movie_files:
+                if os.name == 'nt':
+                    pf_path = pf_path.replace('\\', '/')
```
pdcxs: The ffmpeg seems not able to recognize the backslash in the partial_movie_files_list.txt.

3b1b:
Thanks for this.  I actually want to leave the partial movie files rather than removing them.  The whole motivation for this change is that in video editing, it is easier to work with the smaller pieces, which effectively have cuts at all the points one might want.

Perhaps it's worth adding an option for users that want these removed, so as to not take up twice as much disk space.

Issue #446, opened 2019-02-18
Some shape constructors not working


jeanimal:
I have been getting errors in shape constructors when trying to run the tutorial at https://github.com/zimmermant/manim_tutorial/blob/master/manim_tutorial_P37.py.

I run it like this:
python3 -m manim manim_tutorial_P37.py MoreShapes -pl

First I got an error in the Ellipse constructor: "name 'width' is not defined"-- the full stack is below.  I think I fixed it with
https://github.com/3b1b/manim/compare/master...jeanimal:patch-1

Then I got an error in the CurvedArrow constructor: 'CurvedArrow' object has no attribute 'get_length'
I do not have a proposed fix (which would require implementing get_length()).

Stack trace for Ellipse constructor problem...
Traceback (most recent call last):
  File "/Users/me/Code/python3/manim/manimlib/extract_scene.py", line 153, in main
    scene = SceneClass(**scene_kwargs)
  File "/Users/me/Code/python3/manim/manimlib/scene/scene.py", line 52, in __init__
    self.construct()
  File "manim_tutorial_P37.py", line 33, in construct
    ellipse=Ellipse(width=3, height=1, color=RED)
  File "/Users/me/Code/python3/manim/manimlib/mobject/geometry.py", line 319, in __init__
    self.set_width(width, stretch=True)
NameError: name 'width' is not defined

Stack trace for CurvedArrow constructor problem...
Traceback (most recent call last):
File "/Users/me/Code/python3/manim/manimlib/extract_scene.py", line 153, in main
scene = SceneClass(**scene_kwargs)
File "/Users/me/Code/python3/manim/manimlib/scene/scene.py", line 52, in **init**
self.construct()
File "manim_tutorial_P37.py", line 35, in construct
pointer = CurvedArrow(2_RIGHT,5_RIGHT,color=MAROON_C)
File "/Users/me/Code/python3/manim/manimlib/mobject/geometry.py", line 250, in **init**
self.add_tip()
File "/Users/me/Code/python3/manim/manimlib/mobject/geometry.py", line 45, in add_tip
self.reset_endpoints_based_on_tip(tip, at_start)
File "/Users/me/Code/python3/manim/manimlib/mobject/geometry.py", line 84, in reset_endpoints_based_on_tip
if self.get_length() == 0:
AttributeError: 'CurvedArrow' object has no attribute 'get_length'

jeanimal:
[...]

PR #454, opened 2019-02-24
Move get_length() into TipableVMobject

jeanimal:
This reverts the commit "Cleaned Up Arrow Implementation (d489b5df5c77...), which introduced a bug in the CurvedArrow constructor (https://github.com/3b1b/manim/issues/446).  I had to fix one conflict in the imports, making sure to import both libraries rather than only one.

Done with
git merge upstream/master

https://github.com/3b1b/manim/issues/446

Since I did not see an obvious fix and no one had addressed the bug in several days, I created a rollback now rather than risk more conflicts as other commits are made on top of the bug.

I used https://github.com/zimmermant/manim_tutorial/blob/master/manim_tutorial_P37.py

BEFORE (at head):
python3 -m manim manim_tutorial_P37.py MoreShapes -pl
<snip>
AttributeError: 'CurvedArrow' object has no attribute 'get_length'

AFTER (after revert):
python3 -m manim manim_tutorial_P37.py MoreShapes -pl
<animation is generated>

### DIFF
--- a/manimlib/mobject/geometry.py
+++ b/manimlib/mobject/geometry.py
@@ -12,7 +12,6 @@ from manimlib.utils.iterables import adjacent_pairs
 from manimlib.utils.simple_functions import fdiv
 from manimlib.utils.space_ops import angle_of_vector
 from manimlib.utils.space_ops import angle_between_vectors
-from manimlib.utils.space_ops import center_of_mass
 from manimlib.utils.space_ops import compass_directions
 from manimlib.utils.space_ops import line_intersection
 from manimlib.utils.space_ops import get_norm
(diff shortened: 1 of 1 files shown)


eulertour:
I don't have the time to check this PR now, but as a heads-up the tests currently don't matter at all.

jeanimal:
Do you have time to check this pull request?  It's a very straightforward change generated by
git revert d489b5df5c77a597a71c4fd89597e3d48d5b4d9a
And then merging a line in imports so that both imports still happened.

By "test" I think you mean the travis build, which has been broken for a while: https://travis-ci.org/3b1b/manim/builds

I just synced to head because I saw there were a few more commits and confirmed this still works:
python3 -m manim manim_tutorial_P37.py MoreShapes -pl

That's my version of "tests."

eulertour:
Sorry for being so slow on this. It's unsafe to revert a commit without knowing its motivation, as it likely causes some desired behavior in addition to the bug. It'd be safest to find a way to fix the bug without reverting the entire commit.

jeanimal:
Thanks for taking a look.

We got lucky-- while typing up how I had tried and failed to fix the bug, I saw a simple fix-- moving the get_length() method to the base class.  Please take a look.

I tested it with the tutorial I mentioned before, which generates several shapes, and they all seemed fine:
python3 -m manim manim_tutorial_P37.py Shapes -pl
python3 -m manim manim_tutorial_P37.py MoreShapes -pl

P.S. I don't actually use CurvedArrow in anything I do, so if this request doesn't work out, I'm done.

eulertour:
This looks good. Thanks!

jeanimal:
Thanks for updating the title after the changes.  I deleted the branch but will keep my fork for future bug fixes.

PR #458, opened 2019-02-26
Fixed `temp_file_path` to be more specific in adding the "_temp." string to file name

rodrigo-castellon:
I kept getting an error:
`/Users/rodrigo_temp.castellon/Desktop/misc/pi-day/manim/media/videos/example_scenes/480p15/partial_movie_files/SquareToCircle/00000_temp.mp4: No such file or directory`

Notice how this path doesn't actually exist since `rodrigo_temp.castellon` is not a valid directory, it used to be `rodrigo.castellon`.

So, the code used to produce the temp file path didn't take into account the possibility of paths including periods in other places other than right before the file extension.

It completely resolved the issue I was having -- manim works perfectly now for me -- and seems to have no other side effects (it's just one line that changes).

### DIFF
--- a/manimlib/scene/scene_file_writer.py
+++ b/manimlib/scene/scene_file_writer.py
@@ -199,7 +199,7 @@ class SceneFileWriter(object):
     
     def open_movie_pipe(self):
         file_path = self.get_next_partial_movie_path()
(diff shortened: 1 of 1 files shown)


eulertour:
This relies on the extension having 3 characters, which isn't always the case (e.g. .mpeg, .qt). It'd be better to add a `get_next_temp_partial_movie_path()` function with adds "_temp" in `os.path.join()`.

PR #469, opened 2019-03-08
Fix alignment of minus sign to numbers

ggarza:
This fixes the alignment of the minus sign to negative numbers.

The `align_to` method was being called with an optional argument `alignment_vec` that did not do anything.

This is a before shot:

53990899-f6e35b00-40ee-11e9-907a-1c1f5eef36db

This is an after shot:

Screen Shot 2019-03-08 at 2 46 38 PM

### DIFF
--- a/manimlib/mobject/numbers.py
+++ b/manimlib/mobject/numbers.py
@@ -62,7 +62,8 @@ class DecimalNumber(VMobject):
         # to the bottom
         for i, c in enumerate(num_string):
             if c == "-" and len(num_string) > i + 1:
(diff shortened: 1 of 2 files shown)


ggarza:
Whoops. I only meant to include one commit.  The second commit fixes the position of labels on the y-axis by moving them to the left of the y-axis

3b1b:
Weird, I'm not sure why this hadn't surfaced before.  Or more realistically, what recent change caused it.  In either case, thanks!

PR #495, opened 2019-04-08
Use string multiplication to simplify concatenation loops

Rosuav:
Purely a mechanical change, should have no visible effect whatsoever. Tested by running a quick animation but I have no guarantee that it actually exercised this code.

### DIFF
--- a/manimlib/mobject/svg/tex_mobject.py
+++ b/manimlib/mobject/svg/tex_mobject.py
@@ -108,12 +108,8 @@ class SingleStringTexMobject(SVGMobject):
             tex.count(char)
             for char in "{}"
         ]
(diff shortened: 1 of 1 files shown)


eulertour:
> Tested by running a quick animation but I have no guarantee that it actually exercised this code.

Well, that's not very convincing! But that aside, I don't think the decrease in readability caused by this change is worth the efficiency gain. I do think it should be rewritten, but a simple `if/else` would be easier to read.

Rosuav:
Sure. It's perfectly readable if you're familiar with the idiom of string multiplication, but if you prefer the while loops, no problem, doesn't hurt to create a PR and have it rejected.

PR #500, opened 2019-04-11
Ode

3b1b:
Latest updates while animating the heat equation video.


eulertour:
For what it's worth, I don't know either.

PR #504, opened 2019-04-14
Add files via upload

way2key:
Hi,

I added some installation guide for linux kernel based system...


eulertour:
It seems like the instructions say to download latex twice, once from https://miktex.org/download and a second time via apt. Did you mean to say to add the miktex repository to the users sources? Also, step 3 should have an "s" at the end.

Kolloom:
Documentation to linux debian variant is added. Thanks for the patch

TonyCrane:
Seems outdated, so closed.

PR #508, opened 2019-04-21
Use ==/!= to compare str, bytes, and int literals

cclauss:
Output: https://travis-ci.org/3b1b/manim/builds/522690304#L252


eulertour:
I've really been slacking on getting real tests working, so this should suffice while until I get around to it.

PR #513, opened 2019-04-27
smarter parameterized curve

dor2727:

You should also include:

I wanted to draw a parameterized curve whose 't' goes from -1 to plus infinity

yep


3b1b:
Looks good, thanks!

Issue #515, opened 2019-04-28
Document how to use `stage_scenes.py` (e.g. example)


JohnAZoidberg:
I can't figure out what `stage_scenes` does or how to use it - I'd appreciate some documentation or at least a one-line example. :)

I'm trying to package manim and I'm trying to figure out which files should be runnable. `manim.py` and `stage_scenes.py` seem to be the only, right?

3b1b:
At the moment, I think it's broken, and I wouldn't worry about it too much.  It's something which is meant to organize all the scenes generated from a given file in your file system so that as you move them to a video editor they automatically go in in the correct order.

Arguably it shouldn't be in this repo, but it's not too far off from being more generally useful if I were to go in and clean it up a bit.

PR #516, opened 2019-04-28
Make manim installable using setuptools (setup.py)

JohnAZoidberg:
I want to package manim for my distribution and made changes here so that it's easier for me and others to package it.

Now the project supports the standard Python packaging method of setuptools and we can install it using just `python3 -m pip install .` (even without a distribution packaging it). After installing we can use it in any directory - not just the source repository.

I'm still wondering which are the files that people are supposed to run.
`manim.py` is obviously the main one but does it allow access to all the functionality?
`stage_scenes.py` looks like it does something useful, but I can't figure out exactly what. See: #515 
`manimlib/extract_scene.py` has an `if __name__ == "__main__"` but that's broken because it doesn't call `main()` with all its parameters.

If you like the idea of it being packaged in distribution's package managers I'd also appreciate if you made a release on GitHub with a version number.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,25 @@
+from setuptools import setup, find_namespace_packages
+
+
+setup(name='manim',
+      version='0.1.0',
+      description='Animation engine for explanatory math videos',
+      author='Grant Sanderson',
+      author_email='grant@3blue1brown.com',
+      url='https://github.com/3b1b/manim',
+      license='MIT',
+      packages=find_namespace_packages(),
+      install_requires=[
+         'colour',
+         'numpy',
+         'Pillow',
+         'progressbar',
+         'scipy',
+         'tqdm',
+         'opencv-python',
+         'pycairo',
+         'pydub',
+      ],
+      scripts=['manim.py', 'stage_scenes.py', 'big_ol_pile_of_manim_imports.py'],
+      package_data={'manimlib': ['*.tex', 'files/**']},
+)
--- a/README.md
+++ b/README.md
@@ -5,17 +5,30 @@
 Manim is an animation engine for explanatory math videos. It's used to create precise animations programmatically, as seen in the videos at [3Blue1Brown](https://www.3blue1brown.com/).
 
 ## Installation
-Manim runs on python 3.7. You can install the python requirements with
+Manim runs on python 3.7. You can install the Python requirements with
 `python3 -m pip install -r requirements.txt`. System requirements are
-[cairo](https://www.cairographics.org), [latex](https://www.latex-project.org),
-[ffmpeg](https://www.ffmpeg.org), and [sox](http://sox.sourceforge.net).
+[cairo](https://www.cairographics.org), [ffmpeg](https://www.ffmpeg.org), [sox](http://sox.sourceforge.net), [latex](https://www.latex-project.org) (optional, if you want to use LaTeX).
 
 ### Directly
+Clone this repository and in that directory execute:
 ```sh
-git clone https://github.com/3b1b/manim.git
-cd manim
+# Install python requirements
 python3 -m pip install -r requirements.txt
+
+# Try it out
 python3 -m manim example_scenes.py SquareToCircle -pl
+```
+
+### Installing into your system
+For the previous "direct" method you always have to have this git repository. alternatively you can install it permanently in your system and run it on your own scene files:
+
+```sh
+# Inside repository - Install manim (also installs Python requirements)
+python3 -m pip install .
+
+# Now you don't need the repository anymore and can run it in other directories
+cd anywhere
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
```
FRidh: setup.py is typically not executed, so remove the shebang

JohnAZoidberg: Yeah I guess you're right but they have it in their example: https://docs.python.org/3.7/distutils/setupscript.html
Kenneth Reitz has it too, in his tutorial on `setup.py`: https://github.com/kennethreitz/setup.py/blob/master/setup.py

FRidh: ugh...this makes the assumption `python` exists, which isn't always the case (e.g. when only python3 is around an no symlink exists, which is how a certain PEP describes how it should be)

eulertour:
This is a good idea for manim's development into a real library. A few things to note in order to avoid problems later:

1. Empty `__init__.py` files are redundant in python3.3+, so 5369843 should be reverted. The call to `find_packages` should be replaced with `find_namespace_packages`.
2. `big_ol_pile_of_manim_imports.py` should be added to `scripts`. Really, `big_ol_pile_of_manim_imports.py` should be moved into the library as `manimlib.imports`, but that would be a _very_ breaking change, so it will require at least notifying @3b1b.
3. The version should be 0.1 if not 0.0.1. We don't want to give the impression that this is a stable library yet, and when we do start cutting releases we'll probably end up going through version numbers very quickly.

Regarding your question, everything related to creating video should be accessible through `manim.py`. That said, I don't know what `stage_scenes.py` does.

JohnAZoidberg:
>     1. Empty `__init__.py` files are redundant in python3.3+, so [5369843](https://github.com/3b1b/manim/commit/5369843667f32e0d3a882ae0d98894ecb956eca5) should be reverted. The call to `find_packages` should be replaced with `find_namespace_packages`.

Oh thanks, didn't know that!

>     2. `big_ol_pile_of_manim_imports.py` should be added to `scripts`. Really, `big_ol_pile_of_manim_imports.py` should be moved into the library as `manimlib.imports`, but that would be a _very_ breaking change, so it will require at least notifying @3b1b.

Yes, it should be a part of `manimlib` and yes that's quite breaking therefore I didn't attempt to do it.

I'm not sure about adding it to `scripts` - it's not really a script and it doesn't help to have that installed as a script (or does that make it importable? I'm not sure).
I thought of it more like an example file that you can copy to your project if you wanted to.
In the package I'm writing for my distro I add it do the `PYTHONPATH` would that also happen in `scripts`?

Oh and `manim.py` should probably also be in `manimlib` because that's the recommended structure for Python packages - to have everything in a package.

>     3. The version should be 0.1 if not 0.0.1. We don't want to give the impression that this is a stable library yet, and when we do start cutting releases we'll probably end up going through version numbers very quickly.

Yup. Seems like someone already stole the `manim` name on [Pypi](https://pypi.org/project/manim/) but I'm sure you could get that one because they're not really using it.

eulertour:
> I'm not sure about adding it to scripts - it's not really a script and it doesn't help to have that installed as a script (or does that make it importable? I'm not sure).

I went through the install in a docker container and `big_ol_....` wasn't importable unless it was either moved into the library or added to scripts. I agree that it seems pretty hacky, but then again having one file separate from the rest of the library is pretty hacky in itself. There's a good chance I just don't know enough about setuptools to do it the "right" way.

> Oh and manim.py should probably also be in manimlib because that's the recommended structure for Python packages - to have everything in a package.

It seems that you have some prior experience with setuptools, or have at least done some research into  it. Can you link a source for this recommendation, or any other best practices you know of?

JohnAZoidberg:
> I went through the install in a docker container and `big_ol_....` wasn't importable unless it was either moved into the library or added to scripts. I agree that it seems pretty hacky, but then again having one file separate from the rest of the library is pretty hacky in itself. There's a good chance I just don't know enough about setuptools to do it the "right" way.

Yes, tried it too, works :)

> > Oh and manim.py should probably also be in manimlib because that's the recommended structure for Python packages - to have everything in a package.
> 
> It seems that you have some prior experience with setuptools, or have at least done some research into it. Can you link a source for this recommendation, or any other best practices you know of?

Everyone still always recommends `__init__.py` but `manim` doesn't try to support old Python versions, right?
See the [documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation) of setuptools for why `entry_points` if preferable to `scripts`.
That already assumes that the script is contained in some module. You cannot specify just a file, only a method in a module.
Also see: https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

PR #529, opened 2019-05-08
add option -i to save as gif

Charlsyy:
I added the option to save as gif, however the mp4 or mov file still needs to be saved as well. I'll be trying to find a workaround, but despite that it works. When using the -i option a .gif file will be saved to the output directory

### DIFF
--- a/manimlib/scene/scene_file_writer.py
+++ b/manimlib/scene/scene_file_writer.py
@@ -27,6 +27,7 @@ class SceneFileWriter(object):
         "png_mode": "RGBA",
         "save_last_frame": False,
         "movie_file_extension": ".mp4",
+        "gif_file_extension": ".gif",
         "livestreaming": False,
         "to_twitch": False,
         "twitch_key": None,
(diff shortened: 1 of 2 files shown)


manimlib/scene/scene_file_writer.py
```diff
@@ -299,10 +306,14 @@ def combine_movie_files(self):
             '-f', 'concat',
             '-safe', '0',
             '-i', file_list,
-            '-c', 'copy',
+            #'-c', 'copy',
```
eulertour: You should do this conditionally to avoid breaking changes.

eulertour:
It seems strange that changing the extension alone should create create a .gif. Can you link a source that confirms this?

manimlib/scene/scene_file_writer.py
Charlsyy: Apparently ffmpeg can't can't convert to .gif with this option enabled, that's why I removed it.

Charlsyy:
> It seems strange that changing the extension alone should create create a .gif. Can you link a source that confirms this?

Ffmpeg does exactly that. It recognises the extension and converts it into that format
I also tested that code and it worked just fine. Here is one article showing how to create .gif files with ffmpeg 

https://engineering.giphy.com/how-to-make-gifs-with-ffmpeg/

manimlib/scene/scene_file_writer.py
eulertour: That's fine, but you should only remove it if the `-i` flag was passed.

Charlsyy: Sure, I'll add that as soon as possible

kmein:
Is there something still to be done for this PR to be merged?

eulertour:
Nope, that reminder was enough.

PR #562, opened 2019-05-29
add master_doc to conf.py

eulertour:
A small change, mostly to see under what conditions travis doesn't start a new build.


eulertour:
Works if merged via a merge commit.

PR #573, opened 2019-06-02
update dockerfile, add flag for high quality rendering

eulertour:
The dockerfile now installs manim directly into the container, so there is no need to bind mount the library when running it. That said, you will still have to bind mount the places to read the source and write the video.
1080p rendering is now possible by passing the `--high_quality` flag, and I changed the 1080p frame rate from 30fps to 60fps.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,12 +3,12 @@ RUN apt-get update \
     && apt-get install -qqy --no-install-recommends \
         apt-utils \
         ffmpeg \
-        texlive-latex-base \
         texlive-full \
-        texlive-fonts-extra \
         sox \
         libcairo2-dev \
     && rm -rf /var/lib/apt/lists/*
-COPY requirements.txt requirements.txt
-RUN python -m pip install -r requirements.txt && rm requirements.txt
+COPY . /manim
+RUN cd /manim \
+    && python setup.py sdist \
+    && python -m pip install dist/manimlib*
 ENTRYPOINT ["/bin/bash"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,13 +6,13 @@ services:
     image: eulertour/manim:latest
     # uncomment this line to build rather than pull the image
     # build: .
-    volumes:
-      - ${MANIM_PATH:?MANIM_PATH environment variable isn't set}:/opt/manim
-    environment:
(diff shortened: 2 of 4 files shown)


Kolloom:
I will update the installation documentation when this is merged. So the usage becomes ``docker-compose run manim example_scenes.py SquareToCircle -l`` then?

Kolloom:
I documented the exact dependency based on the tex_template file

``texlive texlive-latex-extra texlive-fonts-extra
texlive-latex-recommended texlive-science texlive-fonts-extra tipa``

Those are the ubuntu packages, should work on debian too. 

I am not sure if this actually saves any space, or if tex_template would change in the future.

eulertour:
> So the usage becomes docker-compose run manim example_scenes.py SquareToCircle -l then?

Yes, as long as the `INPUT_DIR` and `OUTPUT_DIR` environment variables are set. I just updated `docker-compose.yml` to notify the user if they aren't.

eulertour:
> I documented the exact dependency based on the tex_template file
> 
> `texlive texlive-latex-extra texlive-fonts-extra texlive-latex-recommended texlive-science texlive-fonts-extra tipa`
> 
> Those are the ubuntu packages, should work on debian too.

Those packages are included with `texlive-full`, as you can verify [here](https://packages.ubuntu.com/disco/texlive-full).

Kolloom:
I was trying to say using those packages results in smaller download size while satisfies all manim's dependencies. Which might be beneficial to end users.

apt install texlive-full
`0 upgraded, 470 newly installed, 0 to remove and 48 not upgraded.
Need to get 2359 MB of archives.
After this operation, 4511 MB of additional disk space will be used.`

apt install texlive texlive-latex-extra texlive-fonts-extra texlive-latex-recommended texlive-science texlive-fonts-extra tipa
`0 upgraded, 169 newly installed, 0 to remove and 48 not upgraded.
Need to get 1407 MB of archives.
After this operation, 2486 MB of additional disk space will be used.`

Tested on a debian 9 container.

eulertour:
Ah I see, and I do think 2GB is a significant enough difference to warrant the switch.

eulertour:
75996a618c73dfb30b8ca7f69c7266f1dd92571e

PR #582, opened 2019-06-09
Documentation for TipableVMobject

stephenwild:
Add an overview of TipableVMobject's functionality and a few comments, as well as grouping some methods by functionality (roughly).
---------------------------------------------

Just a bit of documentation :) I'm trying to go through manim's source code and see how it all fits together.

As you'd expect, given that I'd only added a few multi-line comments and grouped a few methods together, the scenes in example_scenes.py rendered correctly w/o any dramas.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ ben_cairo_test.py
 .floo
 .flooignore
 .vscode
+.vs
 *.xml
 *.iml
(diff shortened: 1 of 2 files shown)


eulertour:
The docs currently don't display docstrings, but they probably should eventually.

PR #585, opened 2019-06-10
Add documentation for the coordinate system

Kolloom:
Update index
Add 4 assets, one of which is an image

Image uses width 700
Videos uses 700x394
Those dimension are similar to how content is laid out
in sphinx


Kolloom:
those animations are getting quite large, perhaps we should down scale them. 
I currently uses the default which is 1440p. 

1080p is more then enough, plus the fact that the video seems pretty clear even without going fullscreen. Maybe medium setting would work ?

docs/source/coordinate.rst
```diff
@@ -0,0 +1,178 @@
...
+              self.remove(anno)
+
+.. raw:: html
+
+    <video width="700" height="394" controls>
+        <source src="_static/coordinate/DotMap.mp4" type="video/mp4">
+    </video>
+
+.. note::
+  You can place objects outside this boundary, just that it won't show up in the render.
```
eulertour: "but it won't show up in the render."

docs/source/coordinate.rst
```diff
@@ -0,0 +1,178 @@
+Coordinate
+==========
+
+The scene in manim is made up by 8 x 14 grid. The grid is addressed using a numpy
```
eulertour: "By default, the scene..."

docs/source/coordinate.rst
```diff
@@ -0,0 +1,178 @@
...
+        <source src="_static/coordinate/DotMap.mp4" type="video/mp4">
+    </video>
+
+.. note::
+  You can place objects outside this boundary, just that it won't show up in the render.
+
+Using Coordinates
+-----------------
+
+Coordinates are used for creating Geometries and animations.
```
eulertour: There's no object called a Geometry, so this should be reworded.

docs/source/coordinate.rst
```diff
@@ -0,0 +1,178 @@
...
+              "UR": UR,
+              "np.array([1,1,0])": np.array([1, 1, 0]),
+              "DR": DR,
+              "np.array([1,-1,0])": np.array([1, -1, 0])}
+          circle = Circle(color=RED, radius=0.5)
+          self.add(circle)
+          self.wait(0.5)
+
+          for text, aliase in aliases.items():
+              anno = TexMobject(text)
```
eulertour: It'd be better if you can make this monospace rather than italics, but it's not a big deal

docs/source/coordinate.rst
```diff
@@ -0,0 +1,178 @@
...
+              "UP + RIGHT / 2": UP + RIGHT / 2,
+              "DOWN + LEFT * 2": DOWN + LEFT * 2,
+              "RIGHT * 3.75 * DOWN": RIGHT * 3.75 * DOWN,
+              # certain arithmetic won't work as you expected
+              # In [4]: RIGHT * 3.75 * DOWN
+              # Out[4]: array([ 0., -0.,  0.])
+              "RIGHT * 3.75 + DOWN": RIGHT * 3.75 + DOWN}
+
+          for text, aliase in aliases.items():
+              anno = TexMobject(text)
```
eulertour: Same thing about monospace if possible

eulertour:
Also, the medium setting is fine for the docs.

docs/source/coordinate.rst
Kolloom: I didn't know that. Is the dimension dictated by the quality setting?

docs/source/coordinate.rst
Kolloom: I took that from reading geometry.py, not sure people would understand what VMobject is.

docs/source/coordinate.rst
Kolloom: TexMobject defaults to italic, how do I do that?

Kolloom:
I will downscale the assets in a separate patch.

docs/source/coordinate.rst
Kolloom: Well `f"\\textt{ {text} }"` seems to work but the latex includes single quotes `'UP'`

Kolloom: ah `{{{text}}}` gets rid of the quotes

docs/source/coordinate.rst
Kolloom: done

docs/source/coordinate.rst
eulertour: No, but it's possible to zoom the scene.

docs/source/coordinate.rst
eulertour: Seeing it now, I actually think regular text without italics would have been better, but it's fine for now.

PR #587, opened 2019-06-11
Change fill color and stroke color independently

jerpint:

You should also include:

Motivation: using the `set_color` function would change both the stroke and fill color, which is not always convenient (one may wish to change one but not the other).

In this case, if an argument is not provided, the color is simply not updated.

### DIFF
--- a/manimlib/mobject/types/vectorized_mobject.py
+++ b/manimlib/mobject/types/vectorized_mobject.py
@@ -234,9 +234,9 @@ class VMobject(Mobject):
                 sm1.match_style(sm2)
         return self
 
-    def set_color(self, color, family=True):
-        self.set_fill(color, family=family)
-        self.set_stroke(color, family=family)
+    def set_color(self, fill_color, stroke_color=None, family=True):
+        self.set_fill(fill_color, family=family)
+        self.set_stroke(stroke_color, family=family)
         return self
 
     def set_opacity(self, opacity, family=True):


eulertour:
In order to be consistent with the old behavior, you should have the stroke color default to the fill color if none is specified.

jerpint:
I find the old behaviour misleading, i.e. changing the color forces the change to the fill color and stroke color by default. By having the `stroke_color` default to `None`, the stroke color is simply not updated unless you explicitly state so.

eulertour:
In that case you should create new functions to set the fill color and stroke color separately and we can see if set_color can be deprecated.

jerpint:
Would it be too misleading to name it `set_colors` instead of `set_color`?
```
    def set_colors(self, fill_color=None, stroke_color=None, family=True):
        self.set_fill(fill_color, family=family)
        self.set_stroke(stroke_color, family=family)
        return self
```

3b1b:
If you want to actively think about which aspect of the color you're changing, the best ways are:

mob.set_stroke(color=_____)

or

mob.set_fill(color=______)

These also make very clear in the code what the intention is.  Does that not serve what you're going for?

The thinking behind set_color changing both is that most objects are drawn with fill and no stroke (like TexMobjects) or stroke and no fill (like lines, rectangles, etc.), and if you just want to change its color without thinking about that you can just call set_color.

In fact, it would almost certainly cause frustrations if set_color just changed the fill and not the stroke.  Think of an early user who creates a line, then tries to call "line.change_color(RED)" with no effect.

jerpint:
ah ok that makes more sense. I'm still new to manim and have been using it primarily with Square objects in the context of building grids, so for me strokes and fills are both necessary. I hadn't thought of the use case you mention since I haven't come across it yet, but I can see how it would happen a lot. 

Closing this issue :+1:

PR #595, opened 2019-06-16
Move installation procedures out of the readme

yoshiask:
Some people in the manim Discord server pointed out that the readme contained a little too much information. To de-clutter it, I suggest that the installation procedures be moved out of the readme and into their own file.

There also a few other minor changes, such as changing every reference of ``python3`` and ``pip3`` to simply ``python`` and ``pip``.

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -1,0 +1,83 @@
+# General Installation
+Manim runs on python 3.7. You can install it from PyPI via pip
+
+```sh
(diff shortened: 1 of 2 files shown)


eulertour:
I think this information should be moved into the docs rather than another file, so as not to have ~3 different places with installation instructions (README, INSTALLATION, docs).

yoshiask:
I agree, it should be in the docs.

PR #598, opened 2019-06-17
pycairo error

solomonvimal:
fix for pycairo error based on https://github.com/mesonbuild/meson/issues/2273#issuecomment-420412230

You should also include:


Kolloom:
pkg-config and cairo lib are only needed when the wheel installation from pypi fails. This env export is only specific to macos

TonyCrane:

PR #586, opened 2019-06-11
Install doc

Kolloom:
Env Var used in the compose.yml is INPUT_DIR and OUTPUT_DIR

Clarify the meaning of both variables and output message.
Note on -p not working.

PR #604, opened 2019-06-18
fix chapter docker in README.md

ivyxjc:
There is some error in the instruction of how to run manim in Docker.


Kolloom:
There is already a pr for that #586

eulertour:
This got fixed in #586.

PR #605, opened 2019-06-19
Update common_scenes.py

Sh-Ac:
"specific_patrons" misspelt on line 105


Kolloom:
`active_project` `old_project` are not maintained as part of the manim project though. I would assume `for_3b1b_videos` to be so as well.

TonyCrane:
All of the projects of 3b1b has been removed from this repository, so closed this.

PR #609, opened 2019-06-28
Create an add-on API

yoshiask:
The modifications made here are all related to including a new add-on API for Manim. There are a lot of features that could make Manim more useful, but many of those will be quite niche (such as live streaming support, PowerPoint export, timestamps on exported videos). Instead of bloating the Manim with all of these niche features, add-ons can be created and released separately from Manim. The user can decide which additional features they need and download the relevant add-ons.

The API is not currently documented, because I don't want to push the API and then have to make a bunch of breaking changes immediately after. If you have any questions, you can comment on this thread, shoot me an email, or message me in the Manim Discord.


yoshiask:
Why is the travis-ci check failing? It says that it cannot import addon_helper, but it works fine on my machine... does it have something to do with the fact that it is testing it with the PyPI package?

TonyCrane:
Seems outdated, so closed.

PR #612, opened 2019-07-01
Format with autopep8

aDotInTheVoid:
ran `autopep8 --in-place --recursive .`

# Motivation
All code should be formated consistently and in acordence to [PEP8](https://www.python.org/dev/peps/pep-0008/)

# Testing
No semantic code changes


Kolloom:
I don't think an official style has been determined. 

Any how code formatting might be best enforced via git hooks.

eulertour:
I actually think this is a good idea, but if Grant doesn't adhere to it and it's not enforced by Travis, it won't last long.

aDotInTheVoid:
> I don't think an official style has been determined.

I don't believe their are any other style guides out their. There should be one-- and preferably only one --obvious way to do it.

> if Grant doesn't adhere to it and it's not enforced by Travis, it won't last long.

This is very true. I can only show you the door. You’re the one that has to walk through it.

TonyCrane:
The code of master branch has been updated, so this is outdated. But if you want, you can reopen a new one.

PR #624, opened 2019-07-08
Added more details to making_a_scene

npetrangelo:
I added more detail to the making_a_scene file for the documentation. No change to manimlib so it should still be just as functional as it was.


eulertour:
I don't think this addition is necessary, as it doesn't add much to the section, contains some misinformation (declaring Mobjects doesn't add them to the scene), and could potentially be misleading (Animate isn't a real Animation, but that isn't explained).

Issue #627, opened 2019-07-09
Text rendering problem


xy-23:
### Steps to reproduce
```
from manimlib.imports import *

class Playground(Scene):
    def construct(self):
        text = TextMobject("print('Hello, world!')",
                           tex_to_color_map={'print': YELLOW})
        self.play(FadeIn(text))
```
### The unexpected behavior that occurred
Notice between the 'print' and the '(' there is no space.

Wrong
![wrong](https://user-images.githubusercontent.com/47266984/60874808-2fe13700-a26b-11e9-8042-85041a680656.png)

Right
![right](https://user-images.githubusercontent.com/47266984/60874825-35d71800-a26b-11e9-9d20-2dcc2fe515d5.png)
### Solution
I changed [here](https://github.com/3b1b/manim/blob/master/manimlib/mobject/svg/tex_mobject.py) in line 134

`"arg_separator": " ", --> "arg_separator": "",`

and also commented out [here](https://github.com/3b1b/manim/blob/master/manimlib/mobject/svg/tex_mobject.py) in line 160
```
split_list = [str(x).strip() for x in split_list] --> 
#split_list = [str(x).strip() for x in split_list]
```
I also find some interesting in TexMobject(), no matter how many space I enter, like

`text = TexMobject("print    ('Hello, world!')")`

it always ignored those space.

Elteoremadebeethoven:
Thanks for this!! The reason why TexMobject ignores spaces is because inside the quotes you are writing TeX code, and the TeX code ignores the spaces.

sidhanta1989:
Still i am facing the same problem. Not able to generate spaces between the words.
![SpaceIssue](https://user-images.githubusercontent.com/49337352/101286459-a7c0e180-3810-11eb-88b6-d207e68dbad7.jpg)

PR #628, opened 2019-07-09
Solution of text rendering problem

xy-23:
Relevant issues: #627


eulertour:
This PR contains multiple changes, which will make it harder to track the commit if we're searching for one other than the one in the description, and which means they'll all have to be rolled back if there's a problem with any one of them. Additionally, this change breaks at least the following scene:

```python
class WriteStuff(Scene):
    def construct(self):
        example_tex = TexMobject(
            "1", "\\over", "k^2",
        )
        self.play(Write(example_tex))
        self.wait()
```

Since it ends up with a tex file whose contents are `1\overk^2`. This makes me think it would be better just to know that manim works this way an modify the `CONFIG` on a case-by-case basis.

xy-23:
> This PR contains multiple changes, which will make it harder to track the commit if we're searching for one other than the one in the description, and which means they'll all have to be rolled back if there's a problem with any one of them. Additionally, this change breaks at least the following scene:
> 
> ```python
> class WriteStuff(Scene):
>     def construct(self):
>         example_tex = TexMobject(
>             "1", "\\over", "k^2",
>         )
>         self.play(Write(example_tex))
>         self.wait()
> ```
> 
> Since it ends up with a tex file whose contents are `1\overk^2`. This makes me think it would be better just to know that manim works this way an modify the `CONFIG` on a case-by-case basis.

I fixed the problem you mentioned, I think now it can run without problem.

xy-23:
Would anybody merge this Pr or tell me the problem?

PR #631, opened 2019-07-12
Updated installation for linux

nikhithabhupal2000:
The installation guide for linux is half-way. It doesnot give the main thing that python3 version should be 3.7 for the manim software to work . It took almost aday for me figureout that. So,I added some commands to complete the installation guide .

### DIFF
--- a/docs/source/installation/linux.rst
+++ b/docs/source/installation/linux.rst
@@ -17,6 +17,17 @@ Install manim via pypi::
     # pip3 install manimlib
 
 OR Install manim via the git repository with venv::
+NOTE : Make sure that the python version you are using is python3.7 because manim works only for that version . If not, you will get errors when running the old_projects.If the python3 version is not 3.7 then upgrade your version usimg the following steps:
+     
+sudo apt-get install -y software-properties-common
+sudo add-apt-repository ppa:deadsnakes/ppa
+sudo apt-get update
+sudo apt-get install build-essential libpq-dev libssl-dev openssl libffi-dev zlib1g-dev
+sudo apt-get install python3-pip python3.7-dev
+sudo apt-get install python3.7
+
+Then use python3.7 command for 3.7 version and python3 for pre installed version.
+
 
     $ git clone https://github.com/3b1b/manim
     $ cd manim


Kolloom:
I put together a ubuntu 1604 container. 
the python3 inside this container is of version 3.5

f-string is only available from 3.6 onward.
There are three places that uses f string, changing them to .format is very doable. 
```
manimlib/config.py:174:    print(f"Failed to render scene: {str(e)}")
manimlib/constants.py:36:            f"Media will be written to {MEDIA_DIR + os.sep}. You can change "
manimlib/constants.py:246:STREAMING_URL = f"{STREAMING_PROTOCOL}://{STREAMING_IP}:{STREAMING_PORT}?listen"
```

for compiling pycairo two packages are required `pkg-config and libcairo2-dev`
Other usual dependencies applies. 

Further down I encountered this error 
```
root@test:~/manim# python3 -m manim SquareToCircle 
  File "/usr/lib/python3.5/runpy.py", line 184, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/root/manim/manim.py", line 5, in <module>
    manimlib.main()
  File "/root/manim/manimlib/__init__.py", line 11, in main
    config = manimlib.config.get_configuration(args)
  File "/root/manim/manimlib/config.py", line 185, in get_configuration
    module = get_module(args.file)
  File "/root/manim/manimlib/config.py", line 179, in get_module
    module = importlib.util.module_from_spec(spec)
  File "<frozen importlib._bootstrap>", line 574, in module_from_spec
AttributeError: 'NoneType' object has no attribute 'loader'
```
If this is solvable in python 3.5 that would be great. I would recommend against ppa python unless absolutely necessary.

docs/source/installation/linux.rst
```diff
@@ -17,6 +17,17 @@ Install manim via pypi::
     # pip3 install manimlib
 
 OR Install manim via the git repository with venv::
+NOTE : Make sure that the python version you are using is python3.7 because manim works only for that version . If not, you will get errors when running the old_projects.If the python3 version is not 3.7 then upgrade your version usimg the following steps:
```
eulertour: You'll actually get errors running things in old_projects no matter what.

docs/source/installation/linux.rst
```diff
@@ -17,6 +17,17 @@ Install manim via pypi::
...
+     
+sudo apt-get install -y software-properties-common
```
eulertour: These lines aren't formatted for code.

docs/source/installation/linux.rst
```diff
@@ -17,6 +17,17 @@ Install manim via pypi::
...
+sudo add-apt-repository ppa:deadsnakes/ppa
```
eulertour: This should probably only be given as a suggestion. Paranoid users may prefer downloading the source from python.org.

docs/source/installation/linux.rst
```diff
@@ -17,6 +17,17 @@ Install manim via pypi::
...
+sudo apt-get update
+sudo apt-get install build-essential libpq-dev libssl-dev openssl libffi-dev zlib1g-dev
```
eulertour: Do you know why openssl/libssl is needed?

docs/source/installation/linux.rst
dmgolembiowski: @eulertour Would you be able to specify the types of errors that one would normally be getting from the old_projects? (Disclaimer: Today marks my first day working with the repository and I'm on Debian.) I'm wondering if it's even worth debugging these with [https://documen.tician.de/pudb/](url).)

eulertour: Not really; they were written for an older and buggier version of manim, and some are even written for python 2 rather than 3. There's bound to be a lot of constructs used that don't exist in the current version.

nikhithabhupal2000: @ eulertour by old projects I mean the already exsisting projects like example_scenes.py . I would have specified it before

docs/source/installation/linux.rst
nikhithabhupal2000: Okay I will do it

docs/source/installation/linux.rst
nikhithabhupal2000: Yeah ! So, can I add that statement that this is one of the various methods available or should I also give instructions for downloading from python.org?

eulertour: Just say to use python3.7 and offer both options as suggestions.

docs/source/installation/linux.rst
Jaakkonen: PIP uses HTTPS connection to Python package index by default so that may be the reason.

TonyCrane:

Issue #629, opened 2019-07-10
Archlinux installation


tcztzy:
I created an [AUR package](https://aur.archlinux.org/packages/python-manimlib/) to install manim, it works fine as I tested.

Is there any necessity to create a new pull request to update installation document? I think it might be useful for archlinux users.

Kolloom:
You can commit to this file https://github.com/3b1b/manim/blob/master/docs/source/installation/linux.rst

PR #632, opened 2019-07-12
Arch Linux installation guild

tcztzy:
Related #629


eulertour:
The arch install instructions should be under a separate header, like the ones for ubuntu.

tcztzy:
Updated, separated the content as Arch Linux section

PR #636, opened 2019-07-16
[WIP] add dependency check script

Kolloom:
Just wondering if this is a reasonable approach to check manim dependencies.


eulertour:
This looks like a good start for both a `--check-deps` option as well as a test suite.

Kolloom:
What is sox for? I did not find any calls to the program. I don't have sox installed on windows and manim seems running fine.

Kolloom:
My attempt at obtaining a working virtual machine running macos has not been successful. I will not be implementing the macos portion of the script.

eulertour:
@Kolloom the `sox` package contains the `play` binary which is run from `manimlib/utils/sounds.py`.

Kolloom:
@eulertour seems sox is only used when ``--sound`` is set then? This dependency is optional

eulertour:
@Kolloom yeah it's only optional, it's left over from a long time ago.

Kolloom:
patch ready for review

otta8634:
Can you open a new fork, a couple issues with the code.

PR #638, opened 2019-07-18
update windows installtion

Kolloom:
manual wheel install is no longer needed


eulertour:
Why is this not needed anymore? It seems like nobody can get Cairo to work on Windows consistently.

Issue #650, opened 2019-07-26
A Feature Request


xy-23:
### Feature
I added a line of code in \manimlib\extract_scene.py to replace line 119:
`return [scene_classes[0]] if len(scene_classes) == 1 else prompt_user_for_choice(scene_classes)`
This means if your file only has one scene,
you will not to have to declare the scene's name,
the manim will automatically render the scene.

### Motivation
I use vscode to run manim by a line of command like:
`python -m manim ${filebasename} -pl`
So manim can't get the scene's name and I need to type the scene's name by myself, 
which is very annoying.
So I think if I add the feature, I can tap run and get the reasult.
I think this can be a little more convenient for those who just want to test some fuctions
(like beginners).

PR #651, opened 2019-07-26
Add a feature

xy-23:
Relevant issues: #650


eulertour:
The motivation and description should generally go in the PR rather than an issue

PR #652, opened 2019-07-26
Fix tqdm display problem in windows

xy-23:
This is a compromise solution because this problem is caused by windows its own.
You can see this problem in [here](https://github.com/tqdm/tqdm/issues/454).
The problem picture:
problem


xy-23:
Hello？

xy-23:
Anybody here？

eulertour:
Merged

PR #665, opened 2019-08-02
Fixed self.add_sound() for W10 and Community Examples

zavden:
Linux and Mac: `subprocess.call(["rm", sound_file_path])`
Windows: `os.remove(sound_file_path)`

But, this could works in the three OS?


xy-23:
I think it can works in the three os because it is a python-built-in method.
But I only have windows platform to test.

zavden:
Ready @xy-23

eulertour:
Switching to `os.remove()` is a good idea, but community examples shouldn't go in this repo.

zavden:
Ok, let me sended the new one

Issue #444, opened 2019-02-17
i cant run manim successfully


awaqawaq:
i think i've successful installed all module what manim need,but when i run 'python -m manim example_scenes.py SquareToCircle -pl' there were some error i cant solve ...
![image](https://user-images.githubusercontent.com/43958583/52912655-54a23580-32ef-11e9-899f-bb2a5e8b1e0c.png)
how can i solve them thank you very much

Elteoremadebeethoven:
Install readline:
```sh
python3 -m pip install pyreadline
```

awaqawaq:
thanks your reply.
after i installing pyreadline ,i try to run SquareToCircle ,luckily it worked. but when i run other function such as WriteStuff : 
 

![image](https://user-images.githubusercontent.com/43958583/52949228-acf03a80-33b6-11e9-98b9-d98ab65ae4e0.png)


why??

eulertour:
There's a problem with your latex installation, although it's hard to know exactly what it is. You could try reinstalling it.

As for the pyreadline problem, there should be a way to skip the check on Windows. I don't want to install it by default because it seems to be unsupported.

PR #672, opened 2019-08-06
Update requirements.txt to include pyreadline

omertuc:
This solves #444

I never tested this pull request, but I had this error after installing from pip and running the `manim` script:
`ModuleNotFoundError: No module named 'readline'`

I then ran:
`py -m pip install pyreadline`

And got this:
... `Successfully installed pyreadline-2.1`

And now `manim` prints the help message instead of the above error.


VaultDeveloper:
Apparently pyreadline is mandatory for Windows only. Linux users should not download pyreadline instead of readline.

omertuc:
@VaultDeveloper Fixed :+1:

Issue #610, opened 2019-06-29
Latex Error Converting DVI problem


JackDusch:
Include at least:
1. Steps to reproduce the issue (e.g. the command you ran)
2. The unexpected behavior that occurred (e.g. error messages or screenshots)
3. The environment (e.g. operating system and version of manim)

Hello, I have been choogling along with manim and am struggling like many to get the text functional. I managed to implement the changes suggested in #36 which is reiterated in several other threads. I am still able to run commands such as 

```
python3 ./manim.py example_scenes.py SquareToCircle -pl
```

However, I am experiencing an error message relating to the fix involving TEX_DIR that I am unsure how to fix. I tried pointing directly to the Tex directory within manim but didn't think that was consistent with the advice here either. My current tex_file_writing.py looks like:

```
def tex_to_dvi(tex_file):
    result = tex_file.replace(".tex", ".xdv")
    if not os.path.exists(result):
        commands = [
            "xelatex",
            "-no-pdf",
            "-interaction=batchmode",
            "-halt-on-error",
            "-output-directory=" TEX_DIR,
            tex_file,
            ">",
            get_null()
        ]
        exit_code = os.system(" ".join(commands))
        if exit_code != 0:
            log_file = tex_file.replace(".tex", ".log")
            raise Exception(
                "Latex error converting to dvi."
                "See log output above or the log file: %s" % log_file)
    return result   


def dvi_to_svg(dvi_file, regen_if_exists=False):
    """
    Converts a dvi, which potentially has multiple slides, into a
    directory full of enumerated pngs corresponding with these slides.
    Returns a list of PIL Image objects for these images sorted as they
    where in the dvi
    """
    result = dvi_file.replace(".xdv", ".svg")
    if not os.path.exists(result):
        commands = [
            "dvisvgm",
            dvi_file,
            "-n",
            "-v",
            "0",
            "-o",
            result,
            ">",
            get_null()
        ]
        os.system(" ".join(commands))
    return result
```

But, when I attempt to run 
```
python3 ./manim.py example_scenes.py Write Stuff -pl
```
as a test, 

![Screen Shot 2019-06-28 at 10 55 19 PM](https://user-images.githubusercontent.com/51562821/60378926-387a7780-99f8-11e9-94fe-fdf96510e2ad.png)

This is all I can muster. I can still run code not involving text. I tried messing around and removing the TEX_DIR portion, which yields the following message:

[...]

Issue #505, opened 2019-04-16
Exception: Latex error converting to dvi. See log output above...


ostannick:
Let me start by saying that I've read through all of the other issues on here with the same error message, and have tried each of the fixes, but to no avail.

I was running example scene #1, and:

```
Traceback (most recent call last):
  File "C:\Users\kaneo\Desktop\manim\manim\manimlib\extract_scene.py", line 155, in main
    scene = SceneClass(**scene_kwargs)
  File "C:\Users\kaneo\Desktop\manim\manim\manimlib\scene\scene.py", line 52, in __init__
    self.construct()
  File "example_scenes.py", line 18, in construct
    title = TextMobject("This is some \\LaTeX")
  File "C:\Users\kaneo\Desktop\manim\manim\manimlib\mobject\svg\tex_mobject.py", line 145, in __init__
    self, self.arg_separator.join(tex_strings), **kwargs
  File "C:\Users\kaneo\Desktop\manim\manim\manimlib\mobject\svg\tex_mobject.py", line 45, in __init__
    self.template_tex_file_body
  File "C:\Users\kaneo\Desktop\manim\manim\manimlib\utils\tex_file_writing.py", line 19, in tex_to_svg_file
    dvi_file = tex_to_dvi(tex_file)
  File "C:\Users\kaneo\Desktop\manim\manim\manimlib\utils\tex_file_writing.py", line 67, in tex_to_dvi
    "See log output above or the log file: %s" % log_file)
Exception: Latex error converting to dvi. See log output above or the log file: C:\Users\kaneo\Desktop\manim\manim\manimlib\files\Tex\cc8b128e228b77ce.log
```

The 'log file' contains:

`\documentclass[preview]{standalone}

\usepackage[english]{babel}
\usepackage{amsmath}
\usepackage{amssymb}
\usepackage{dsfont}
\usepackage{setspace}
\usepackage{tipa}
\usepackage{relsize}
\usepackage{textcomp}
\usepackage{mathrsfs}
\usepackage{calligra}
\usepackage{wasysym}
\usepackage{ragged2e}
\usepackage{physics}
\usepackage{xcolor}
\usepackage{textcomp}
\usepackage{microtype}
\DisableLigatures{encoding = *, family = * }
%\usepackage[UTF8]{ctex}
\linespread{1}

\begin{document}

\centering This is some \LaTeX

\end{document}`

I'm on windows 10. This same error happens when I make my own custom scripts, but I am able to run example scenes 2 and 4.

[...]

PR #689, opened 2019-08-17
Solve 'Latex error converting to dvi' problem on Windows

chaoxiaosu:
I faced the same problem as [#505](https://github.com/3b1b/manim/issues/505) and [#610](https://github.com/3b1b/manim/issues/610) on my Windows 10 machine. After making this change, I passed through **manim** working on scenes with TeX stuff.


chaoxiaosu:
The problem, I think, comes from that `latex` can not deal with tex file having path like this `./media\\Tex\\xxx.tex`. 

On windows, when `manim` scene without specifying `--media_dir` or `tex_dir` option,  by default, the TEX_DIR defined in L41 of manimlib/constants.py will be `./media\\Tex`. Then the tex file written by _tex_file_writing.py_ will have path like this `./media\\Tex\\xxx.tex`.

zombie110year:
In my opnion, The true reason is https://github.com/3b1b/manim/issues/505#issuecomment-534573115

Issue #688, opened 2019-08-16
References to an object in submobjects breaks across a copy


hexagonal-sun:
Suppose we have the following code:

```python
class Video(Scene):
        a = VGroup()
        a.foo = Circle()
        a.bar = Rectangle()
        a.digest_mobject_attrs()
        a.print_family()
        print(id(a.foo), id(a.bar))
        self.add(a)
        self.wait()

        a.generate_target()
        a.target.remove(a.target.foo)
        a.target.print_family()
        print(id(a.target.foo), id(a.target.bar))

        self.play(MoveToTarget(a))
        self.wait()

```

which results in:

```
 VGroup 139926527884496
	 Circle 139926858136272
	 Rectangle 139926527884688
139926858136272 139926527884688
```
Everying is all good at this point, however after the copy, `a.target.foo` and `a.target.bar` are no longer the same objects on `a.target`'s submobject array
```
 VGroup 139926527885136
	 Circle 139926527885264
	 Rectangle 139926621818128
139926527967696 139926527967632
```
and therefore, the removal of a.target.foo doesn't work.

I think this is a problem [here](https://github.com/3b1b/manim/blob/master/manimlib/mobject/mobject.py#L130), the `value.copy()` results in two objects.

Expected behaviour: the circle should be removed in the animation.

Elteoremadebeethoven:
Sorry, I don't know what you are trying to prove, `MoveToTarget` was not thought that way, if you want to make that animation it is much easier to do it this way (by the way, you forgot to include the `construct` method).

```python
class Video(Scene):
    def construct(self):
            a = VGroup()
            foo = Circle()
            bar = Rectangle()
            a.add(foo,bar)
            self.add(a)
            self.wait()

            self.play(FadeOut(foo))
            self.wait()
```

If you plan to apply this code in a more complex animation I ask you to explain more.

hexagonal-sun:
Apologies, my example wasn't very clear on how I came across this problem.  This is mainly an issue when you are trying to implement a subclass.  I'm creating videos about programming in C and I've subclassed VGroup into a new object, `CObject`.  The example below should hopefully highlight the problem:

```python
from manimlib.imports import *

class CObject(VGroup):
    CONFIG = {
        "label": "CObject",
        "attrs": {}
    }

    def __init__(self, **kwargs):
        super(CObject, self).__init__(**kwargs)
        digest_config(self, kwargs)
        self.label = TextMobject("\\texttt{" + self.label + "}",
[...]

PR #691, opened 2019-08-18
manimlib/mobject/mobject: use deepcopy by default

hexagonal-sun:
Using a shallow copy can result in duplicate objects for references in
an mobject that aren't part of the submobjects array, see [1].
Deepcopy fixes this and doesn't seem to have any other adverse
drawbacks.

Use it by default, however add a config value allowing this to be
overwritten for backwards compatibility.

[1]: https://github.com/3b1b/manim/issues/688

This has been tested by compiling most of the built-in scenes.

### DIFF
--- a/manimlib/mobject/mobject.py
+++ b/manimlib/mobject/mobject.py
@@ -34,6 +34,7 @@ class Mobject(Container):
         "name": None,
         "dim": 3,
         "target": None,
+        "use_deepcopy": True,
     }
 
     def __init__(self, **kwargs):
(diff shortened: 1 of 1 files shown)


yoshiask:
I'd like to point out that the code for `mobject.copy()` even states that it should just `return self.deepcopy()`. https://github.com/3b1b/manim/blob/dbdd7996960ba46ed044a773290b02f17478c760/manimlib/mobject/mobject.py#L117

TonyCrane:
Seems outdated, so closing this :)

PR #703, opened 2019-08-21
Fix a typo in README.md

ShutoAraki:
I just got started with manim, but I found one typo in the documentation so here it is.

The `OUTPUT_PATH` for `docker-compose` creates `videos/` directory instead of `media/` and it is confusing for a beginner.


TonyCrane:
The README file has been rewrote, so this is outdated.

PR #711, opened 2019-08-28
add frame_rate command line argument

tomasderner97:
There is apparently no way to set fps without editing manim source code. This adds a command-line argument frame_rate. I think this is useful, there are many cases where rendering with 60 fps is wasteful and not fitting with the rest of the project outside manim (we use 24 fps).

### DIFF
--- a/manimlib/config.py
+++ b/manimlib/config.py
@@ -103,6 +103,10 @@ def parse_cli():
             help="Background color",
         )
         parser.add_argument(
+            "--frame_rate",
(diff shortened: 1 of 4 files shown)


dor2727:
You may use
parser.add_argument(type=int)
So that argparse will automatically read frame_rate as an integer. no need for manual conversion.

TonyCrane:
The frame_rate arg has already been added in the new version, so closed this.

PR #734, opened 2019-09-24
latex can't recognize path seperator \\ in windows

zombie110year:
Modify path as posix style before transport it to system command to fix issue #505 

I ran the example_scenes.py and pass all cases.

### DIFF
--- a/manimlib/utils/tex_file_writing.py
+++ b/manimlib/utils/tex_file_writing.py
@@ -1,6 +1,8 @@
 import os
 import hashlib
 
+from pathlib import Path
+
 from manimlib.constants import TEX_TEXT_TO_REPLACE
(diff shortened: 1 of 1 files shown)


sajidnisar:
Thanks @zombie110year; a must change for manim to work on Windows (10) with latex!

435vic:
Definitely an issue that should be looked into.

markchalmersUC:
Just wanted to say thank you! this is the only thing that fixed my latex issue on windows 10.

TooYoungTooSimp:
A better fix than my local `.replace('\\','/')` !
It's a must merge fix for manim indeed.

filipppp:
Can we please merge this?

PR #755, opened 2019-10-11
Use generators for any and all calls and comprehensions

tweakimp:
I just made some small improvements:
In `any` and `all` calls, generators are faster than lists because the loop can break and short cut if the first `True` (in `any`) or `False` (in `all`) is found.
Also some cleaner code with comprehensions :)

### DIFF
--- a/active_projects/eop/reusables/histograms.py
+++ b/active_projects/eop/reusables/histograms.py
@@ -2,11 +2,7 @@ from manimlib.imports import *
 from random import *
 
 def text_range(start,stop,step): # a range as a list of strings
-    numbers = np.arange(start,stop,step)
-    labels = []
-    for x in numbers:
-        labels.append(str(x))
-    return labels
+    return [str(x) for x in np.arange(start,stop,step)]
 
 
 class Histogram(VMobject):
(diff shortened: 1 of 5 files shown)


dor2727:
shouldn't you, following the same logic, in the 2 changes in "active_projects/eop/reusables/histograms.py", change the square brackets into parentheses? So that it would return a generator?

tweakimp:
no, because i dont know where they are used. maybe the function that calls them expects them to be lists.

if i know the caller just iterates over the results, i could turn the called functions into generators with the yield statement, but that requires deeper analysis of the code

active_projects/eop/reusables/histograms.py
```diff
@@ -82,10 +78,7 @@ def generate_points(self):
             self.remove(submob)
 
         def empty_string_array(n):
-            arr = []
-            for i in range(n):
-                arr.append("")
-            return arr
+            return ["" for _ in range(n)]
```
miggazElquez: Why not ``[""] * n`` ?

TonyCrane:
Seems outdated, so I'm closing this.

PR #813, opened 2019-12-03
Add plot_depth to introduce layer concept

pdcxs:
I want to draw some graphs beneath the graphs which are already previously created. So I add a plot_depth property to the mobject, higher value means higher layers. So I can create a line under two points like:

```python
p1 = Dot(LEFT)
p2 = Dot(RIGHT)
p1.plot_depth = 1
p2.plot_depth = 1
line = Line(LEFT, RIGHT)
self.play(ShowCreation(p1), ShowCreation(p2)
self.play(ShowCreation(line))
```

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,26 +1,8 @@
-All files of this project under the directories active_projects and old_projects
-are copyright 3Blue1Brown LLC and used by permission for this project only.
+The MIT License (MIT)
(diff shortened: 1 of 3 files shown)


Tony031218:
Here is a bug: https://github.com/ManimCommunity/manim/issues/327

TonyCrane:
Outdated.

PR #815, opened 2019-12-04
Add documentation to manimlib/constants.py

philippeitis:
1. Added documentation specifying how directories are initialized. Tweaked the help messages to be clearer, and specified that animations are 0 indexed. This was inspired by the readme, which specifies that the project needs better documentation.

2. This only adds documentation and tweaks help strings, and does not touch any program logic. You can test this by running code where a font is missing, a scene is missing, or choosing a number is required. 

You should also include:

### DIFF
--- a/manimlib/constants.py
+++ b/manimlib/constants.py
@@ -9,12 +9,23 @@ TEXT_DIR = ""
 
 
 def initialize_directories(config):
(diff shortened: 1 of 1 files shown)


TonyCrane:
The constants.py has been updated, so this is outdated, I'm closing this.

PR #818, opened 2019-12-05
Add Physics Engine Support

pdcxs:
I want to play physics engine with manim. So I tried to add a simple wrap of the pymunk library. The test scene is added into the example_scenes.py.

With the help of physics engine, I think this library can be more powerful.


pdcxs:
More examples: https://www.bilibili.com/video/av78234079

Issue #694, opened 2019-08-19
can't graph trigonometric function of secx, cscx, cotx, tanx,...


barakasamsara:
source  code:

class PlotFunctions(GraphScene):
    CONFIG = {
        "x_min" : -10,
        "x_max" : 10.3,
        "y_min" : -1.5,
        "y_max" : 1.5,
        "graph_origin" : ORIGIN ,
        "function_color" : RED ,
        "axes_color" : GREEN,
        "x_labeled_nums" :range(-10,12,2),

    }
    def construct(self):
        self.setup_axes(animate=True)
        func_graph=self.get_graph(self.func_to_graph,self.function_color)
        func_graph2=self.get_graph(self.func_to_graph2)
        vert_line = self.get_vertical_line_to_graph(TAU,func_graph,color=YELLOW)
        graph_lab = self.get_graph_label(func_graph, label = "\\cos(x)")
        graph_lab2=self.get_graph_label(func_graph2,label = "\\sin(x)", x_val=-10, direction=UP/2)
        two_pi = TexMobject("x = 2 \\pi")
        label_coord = self.input_to_graph_point(TAU,func_graph)
        two_pi.next_to(label_coord,RIGHT+UP)

        self.play(ShowCreation(func_graph),ShowCreation(func_graph2))
        self.play(ShowCreation(vert_line), ShowCreation(graph_lab), ShowCreation(graph_lab2),ShowCreation(two_pi))

    def func_to_graph(self,x):
        #return np.cos(x)
        return np.tan(x)

    def func_to_graph2(self,x):
        return np.sin(x)

I replaced "return np.cos(x)" to "return np.tan(x)"...i got this:
![image](https://user-images.githubusercontent.com/36161299/63267544-e140a700-c2c4-11e9-9164-a14d37ee8673.png)

and then I replaced "return np.cos(x)" to "return np.sec(x)/cot(x)/csc(x)"...i got this:
AttributeError: module 'numpy' has no attribute 'sec'...

zavden:
Those functions are discontinuous, you must break them into pieces to make them work, like [https://github.com/3b1b/manim/issues/449]. Also you can import the math package of python instead numpy.

barakasamsara:
> Those functions are discontinuous, you must break them into pieces to make them work, like [this](https://github.com/3b1b/manim/issues/449). Also you can import the math package of python instead numpy.

Thanks very much, but it didn't work... 
i got this:
[...]

Issue #498, opened 2019-04-11
complex functions acting strange


tjf801:
When I try to run a complex transformation, it runs fine, unless the function has a pole. I would like if someone could tell me what I could do to fix this. Thanks in advance.

Code:
![Capture](https://user-images.githubusercontent.com/47402907/55921882-b9319080-5bcc-11e9-8518-f8e60adc8728.PNG)

Middle:
![Capture](https://user-images.githubusercontent.com/47402907/55922112-bedba600-5bcd-11e9-926c-478d17cf63f9.PNG)

Last frame:
![Capture](https://user-images.githubusercontent.com/47402907/55922131-d581fd00-5bcd-11e9-8504-5cb145a37ca7.PNG)

Rousan99:
Can you show me how you get complex Transformation of the whole coordinate plane for any Function?

tjf801:
I'm A little late but you change the lambda function in CONFIG to whatever you want.

Rousan99:
But I don't understand why you use self.apply_complex_function(self.function)
Why not self.apply_complex_function(zeta)?

giladsofer:
Hi, were you able to solve this problem? I'm having the same problem myself right not, wondering if there's some way to fix it. Thank you

tjf801:
sadly no. I wish though lol

giladsofer:
Right now I'm trying to check the `[discontinuity fix code](https://github.com/3b1b/manim/pull/832)`, but I'm having some trouble understanding how to use it.

yinjann:
After several days I managed to solve the problem with the current version of manim
#1169
#1170
#1171
![ur_k](https://user-images.githubusercontent.com/42347160/87492476-42c1e080-c610-11ea-9aa3-8f59dab24d2b.png)

giladsofer:
> After several days I managed to solve the problem with the current version of manim
> #1169
> #1170
> #1171
> ![ur_k](https://user-images.githubusercontent.com/42347160/87492476-42c1e080-c610-11ea-9aa3-8f59dab24d2b.png)

Could you please post the code you used for this specific animation? Just to see how your method is applied in context? Thank you

Rousan99:
> After several days I managed to solve the problem with the current version of manim
> #1169
> #1170
> #1171
> ![ur_k](https://user-images.githubusercontent.com/42347160/87492476-42c1e080-c610-11ea-9aa3-8f59dab24d2b.png)

[...]

PR #832, opened 2019-12-16
Add better discontinuity functionality.

ddxtanx:
 I made this change because I wanted to work with some algebraic curves that have multiple infinite discontinuities and, with the current version of manim, animating them lead to very ugly artifacts appearing on the curve. This fundamentally boils down to the `discontinuities` array appearing in the ParametricFunction class that we needed to "be smarter about." While this isn't a perfect solution (as there are still 'double-tangent' artifacts present in all the examples (without and with the new discontinuities function), this is a little bit smarter.
 The image below demonstrates the issue I was attempting to solve.
ParametricFunctionScene_With_Old_Discont
 With the new `get_discontinuities` function I coded, the image becomes
ParametricFunctionScene_With_New_Discont
 And, to test both infinite discontinuities and break discontinuities, I just drew a hyperbola with a step function at the tail end, which looked like:
ParametricFunctionScene_Inf_And_Break_Discont
So the issue is solved, and everything should _still_ work.

### DIFF
--- a/manimlib/mobject/functions.py
+++ b/manimlib/mobject/functions.py
@@ -12,6 +12,8 @@ class ParametricFunction(VMobject):
         "dt": 1e-8,
         # TODO, be smarter about figuring these out?
         "discontinuities": [],
+        "tol_point": 1e5,
+        "tol_del_mult": 1e-2
     }
 
     def __init__(self, function=None, **kwargs):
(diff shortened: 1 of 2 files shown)


ddxtanx:
With the new break discontinuity detection, the tangent lines in the curve now disappear, while still making sure the break in the step function is gone as well.

image
image

ddxtanx:
Latest commit addresses #694 by extending discontinuity detection to GraphScene graphs.

ddxtanx:
And on appropriate choice of tol_point (I set it equal to 15, since this example's y axis only goes to 1.5 and x axis to 10) the graph becomes:
image
which is free of vertical line artifacts.

ddxtanx:
> These discontinuity artifacts also seem to be an issue with complex functions as well.
> 
> Relevant issue: #498

From what I can see in the ComplexFunctionScene code, there are no references to ParametricFunction, so these changes would not, unfortunately, affect complex_function_scene. Seeing as Grant put ComplexFunctionScene in once_useful_constructs and has not been updated or tinkered with in a while, I suspect it might be deprecated.

giladsofer:
I tried running your TestTrigDiscont example but it didn't work (it didn't fix the noise due to discontinuity). Is there something else you did other than using this code? Generally, I don't really understand how exactly I should use tol_point.

TonyCrane:
Seems outdated, so closed this.

PR #841, opened 2019-12-27
Add simple table drawing functionality.

Aathish04:
## Motivation
Tables are really useful for displaying data concisely, which is really useful when making animations on topics such as statistics and science. Currently, Manim can only generate Tables easily through LaTeX, which results in difficulty when generating tables programmatically.

## Results Obtained
With this addition, tables can be made by calling `Table.get_table(<dictionary>)`, where the main, required parameter is a dictionary with the Keys as the field names and Values per Key are the corresponding records.

```python
class Tables(Scene):
    def construct(self):
        tabledict={
            "Vowels":["\\alpha","\\epsilon","\\iota","\\omega","a","e","i","o"],
            "Consonants":["\\beta","\\gamma","\\delta","\\phi","\\eta","\\kappa","\\lambda","\\mu","\\nu","\\pi","\\chi","\\zeta"],
            "Words":["Must","add","proper","length","checking","for","use","with","LaTeX.","(No","wraparound","currently.)"],
        }
        table=Table.get_table(tabledict)
        table.scale(0.50)
        table.move_to((0,0,0))
        self.play(Write(table))
        self.wait(2)

```
The code above generates the following table:

Tables

Further customisations can be made by passing a `line_color` and `text_color` in the same function like so: `Table.get_table(tabledict,line_color=GREEN,text_color=BLUE)`
Both of these values default to WHITE.

## Changes Made:

A single file named `tables.py` has been added to `manimlib/mobjects`. This file contains the entirety of the logic used.

`manimlib.imports` has been edited to import the aforementioned file.

Future plans for this feature include a wraparound for text and ~~better length measuring of LaTeX strings to ensure proper table column widths, as well as~~ better table element retrieval, since you can only retrieve elements by indexing the resultant VGroup() as of now.

EDIT: I'm closing this PR, since I feel this would work better as an addon to Manim rather than as a core feature. I added support for Text(), TexMobjects(), TextMobjects() and strings. Those interested can find it [here](https://github.com/Aathish04/sanim/tree/master/anim_tools). An example on how to use all its features are [here](https://gitlab.com/aathish04/manim-projects/tree/master/CustomFeatureExamples).

### DIFF
--- a/manimlib/imports.py
+++ b/manimlib/imports.py
@@ -58,6 +58,7 @@ from manimlib.mobject.types.vectorized_mobject import *
 from manimlib.mobject.mobject_update_utils import *
 from manimlib.mobject.value_tracker import *
 from manimlib.mobject.vector_field import *
+from manimlib.mobject.tables import *
 
 from manimlib.for_3b1b_videos.common_scenes import *
 from manimlib.for_3b1b_videos.pi_creature import *
--- a/manimlib/mobject/tables.py
+++ b/manimlib/mobject/tables.py
@@ -1,0 +1,95 @@
+from manimlib.mobject.types.vectorized_mobject import VGroup
+from manimlib.mobject.svg.tex_mobject import TexMobject
+from manimlib.mobject.geometry import Line
+from manimlib.constants import *
+class Table(VGroup):
+#TODO : Add proper length checking to avoid smaller than useful vertical lines when dealing with TeX
+# Use CONFIG instead of directly passing parameters to function?
+
+    def get_table(tabledict:dict,buff_length=0.3,line_color=WHITE,text_color=WHITE):
+        
+        def flatten(inlist):
+            outlist=[]
+            for element in inlist:
+                for sub_element in element:
+                    outlist.append(sub_element)
+            return outlist
+        
+        table=VGroup() #The table is a VGroup with all the fields, records and separators.
+        
+        fields=list(tabledict.keys()) #Since the data is recieved as a dict, the keys will be the fields
+
+        cell_length=TexMobject(max(fields + flatten(tabledict.values()), key=len)).get_width()+2*buff_length #The length of a record/field of max length is the base cell size
+        cell_height=TexMobject(max(fields + flatten(tabledict.values()), key=len)).get_height()+2*buff_length
+       
+        #The first position is set like so.
+        field_position=[ (cell_length-TexMobject(fields[0]).get_width())/2 + TexMobject(fields[0]).get_width()/2, 0,0 ] #The initial position of the first field. This is 
+        #NOTE: Coordinates of TexMobjects in Manim are taken from centre, not top-right. Adjustments have been made.
+        
+        total_table_width=(cell_length-TexMobject(fields[0]).get_width())/2
+
+        total_table_height=cell_height*(len(tabledict[max(tabledict.keys(),key=len)])+1)
+
+        for n in range(len(fields)):
+            
+            field=TexMobject(fields[n])
+            
+            field_length=field.get_width() #This is the length that the actual field name will take up on-screen
+            
+            if n+1<len(fields): #This gets the nxt field if it exists and chooses an empty TexMobject if it doesn't
+                next_field=TexMobject(fields[n+1])
+            else:
+                next_field=TexMobject("")
+            
+            next_field_length=next_field.get_width() #Gets the next fields length
+            
+            
+            field.move_to(field_position)
+            
+            
+            space_to_right_of_field=(cell_length-field_length)/2
+            
+            space_to_left_of_next_field=(cell_length-next_field_length)/2
+            
+            space_to_leave= space_to_right_of_field + space_to_left_of_next_field + next_field_length/2
+            
+            #next_field_length/2 is added to account for the fact that coordinates are taken from centre and not left edges.
+
+            total_table_width+=field_length+space_to_leave/2
+            table.add(field)
+            field_position=field.get_right()+(space_to_leave,0,0)
+        
+        for keynum in range(len(tabledict.keys())):
+            key=list(tabledict.keys())[keynum] #gets the actual key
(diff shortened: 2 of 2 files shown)


naths84:
Hi thanks for this great work. Can you update the table to add info in it after it has been created?

Aathish04:
> Hi thanks for this great work. Can you update the table to add info in it after it has been created?

I'm still working on that. However, as a workaround, you can index the table like a list, since it's just a VGroup.

The way I've set it up now, the first few objects in the VGroup are the field names/headings.

After the field names, the next few objects are reserved for those in the first column (the leftmost one) and are ordered top to bottom.

The next few objects are the values in the column to the immediate right of it and so on, until all the records are in the VGroup. 

The object present immediately after the records is the horizontal line. The objects present after that are the vertical separators. 

Once you know which object to modify, you can ReplacementTransform() them with the new value you want them to be.

For example:
```python
from sanim.anim_tools.tables import *
class Tables(Scene):
    def construct(self):
        
        tabledict={
            TextMobject("TextMobject Input"):[TextMobject("Must"),TextMobject("add"),TextMobject("element"),TextMobject("retrieval.")],
            TexMobject("TexMobject Input"):[TexMobject(r"e^{\iota\pi}+1 = 0"),TexMobject(r"Tex: \alpha\theta\epsilon")],
            Text("Text input",font="Lucida Grande"):[Text("Text",font="sans-serif"),Text("is"),Text("Supported")],
            "Raw String Input":["Defaults","to","TextMobject."]

        }

        table=Table.get_table(tabledict,line_color=GREY,raw_string_color=BLUE)
        table.move_to((0,0,0))
        self.play(Write(table.scale(0.5)),run_time=2)
        #CHECK THIS OUT ↓
        newvalue=TextMobject("New Value").move_to(table[0]) #The first index is the first columns heading
        self.play(ReplacementTransform(table[0],newvalue))

        self.wait(1)
```

The code above gives this output:



TableExample

naths84:
Thanks it worked. The counting does not include empty table cells? It's a bit strange. Looking forward to the rest of your code!

Aathish04:
Perhaps you could use empty strings for those locations?

Aathish04:
> Thanks it worked. The counting does not include empty table cells? It's a bit strange. Looking forward to the rest of your code!

Hey! I managed to add basic Record addition and removal to the table.

You can get the most recent version of the script here: https://github.com/Aathish04/sanim/

I've had to make some major changes regarding how everything works.

Table() is now it's own type of Mobject, and not just a VGroup. You can still use it like a VGroup though.

You can use `Table.add_record(record object,field_number)` to add records.
You can use  `Table.remove_record(field_number,record_number)` to remove  the record_numbered record in a field.

See the example below:
```python
from sanim.anim_tools.tables import *
class Tables(Scene):
    def construct(self):
        
        tabledict={
            TextMobject("TextMobject Input"):[TextMobject("Must"),TextMobject("add"),TextMobject("element"),TextMobject("retrieval.")],
            TexMobject("TexMobject Input"):[TexMobject(r"e^{\iota\pi}+1 = 0"),TexMobject(r"Tex: \alpha\theta\epsilon")],
            "Raw String Input":["Defaults","to","TextMobject."],
            Text("Text input",font="Lucida Grande"):[Text("Text",font="Alys Script Bold"),Text("is",font="serif"),Text("Supported",font="serif")],
        }

        table=Table(tabledict=tabledict,line_color=GRAY,raw_string_color=BLUE)
        table.move_to((0,0,0))
        table.scale(0.5)
        
        self.play(Write(table),run_time=2)

        self.play(Write(table.add_record(record=TextMobject("Hello"),field_num=1)))
        
        self.play(Uncreate(table.remove_record(field_num=1,record_num=0)))
        
        self.wait(1)
```
Hope this helps!

naths84:
For the empty string, I had already tried and the cell was ignored in the numbering.

Thanks a lot for the new code, I'll use it asap and let you know if I have any feedback.

Aathish04:
Please do tell me if there are any wierd bugs etc!

aieml:
image

I am getting this error. Any Suggestions?

Aathish04:
First Off:
You’re using a really old version of the Manim Library.

Instead of using the version with Pip, you should use the version from GitHub.

almccutc:
I'm having the error:

    table=Table.get_table(tabledict)
NameError: name 'Table' is not defined

I put the file tables.py into the folder manimlib/mobjects, not sure what the issue is.

Tony031218:
> I'm having the error:
> 
> ```
> table=Table.get_table(tabledict)
> ```
> 
> NameError: name 'Table' is not defined
> 
> I put the file tables.py into the folder manimlib/mobjects, not sure what the issue is.

Have you added `from manimlib.mobjects.tables import *` to `manimlib/imports.py` ?

Aathish04:
> I'm having the error:
> 
> ```
> table=Table.get_table(tabledict)
> ```
> 
> NameError: name 'Table' is not defined
> 
> I put the file tables.py into the folder manimlib/mobjects, not sure what the issue is.
@almccutc 

Like @Tony031218 mentioned, also add `from manimlib.mobjects.tables import * to manimlib/imports.py`, so you can import it that way.

Also, `Table.get_table(<dict>)` has been deprecated, and you should use `Table(<dict>)` instead.

Aathish04:
FOR THE REFERENCE OF ANYONE WHO STUMBLES UPON THIS THREAD:
Please use https://github.com/Aathish04/sanim/blob/manim-3b1b/anim_tools/tables.py for the most recent, bug free version of this feature.

zdarovakoresh:
Simple adding the latest version of `tables.py` doesn't work. But the  previous version (95 lines) works good.

Aathish04:
> Simple adding the latest version of `tables.py` doesn't work. But the  previous version (95 lines) works good.

Can you please file an issue on the repository I linked stating what went wrong and the traceback/errors, if any?

Issue #845, opened 2019-12-31
BrokenPipeError Mac


mattdeitke:
Whenever I try rendering clips > ~30 seconds, I occasionally get the **BrokenPipeError** bug:
```
Traceback (most recent call last):
  File "~/Dropbox/documents/code/manim/manim/manimlib/extract_scene.py", line 155, in main
    scene = SceneClass(**scene_kwargs)
  File "~/Dropbox/documents/code/manim/manim/manimlib/scene/scene.py", line 53, in __init__
    self.construct()
  File "~/Dropbox/documents/code/manim/manim/active.py", line 147, in construct
    self.wait(2.5)
  File "~/Dropbox/documents/code/manim/manim/manimlib/scene/scene.py", line 406, in wrapper
    func(self, *args, **kwargs)
  File "~/Dropbox/documents/code/manim/manim/manimlib/scene/scene.py", line 521, in wait
    self.add_frames(*[frame] * n_frames)
  File "~/Dropbox/documents/code/manim/manim/manimlib/scene/scene.py", line 543, in add_frames
    self.file_writer.write_frame(frame)
  File "~/Dropbox/documents/code/manim/manim/manimlib/scene/scene_file_writer.py", line 185, in write_frame
    self.writing_process.stdin.write(frame.tostring())
BrokenPipeError: [Errno 32] Broken pipe
Media will be written to ./media/. You can change this behavior with the --media_dir flag.
```

This bug does not always appear, and I'd estimate that it appears about 80% of the time at different points throughout the rendering process.

This bug seems different from all the previous BrokenPipeError's reported, and I'm really not sure what's causing it. It seems to only happen when there are a lot of animations. This has nothing to do with #669 and I'm not using a Conda environment (as in #536  + #492).

Any ideas what I can do?

mattdeitke:
I am able to fairly easily reproduce this by creating a new virtualenv and running all the install commands listed [here](http://bhowell4.com/manic-install-tutorial-for-mac/) followed by
```
python3 manim.py example_scenes.py OpeningManimExample -pl
```
Which seems to be the longest example scene. Note: the other example scenes work (I suspect because they are shorter?)

mattdeitke:
Hacky solution, but I seemed to have solved everything.

The bottleneck seems to be on line ~460 in manimlib/scene/scene.py:
```
self.progress_through_animations(animations)
```

Replacing that line with
```
while True:
            try:
                self.progress_through_animations(animations)
                break
            except:
[...]

PR #846, opened 2020-01-01
Fix BrokenPipeError caused by FFmpeg writing to closed stdin

mattdeitke:
When working with relatively large images in manim on my Mac, a BrokenPipeError occurs when FFmpeg tries to write to `file_writer` after stdin is closed.

The error shows up at seemingly random times, even with the same input. For instance, the output below shows that the error occurs at 3 different locations with the exact same input.
Screenshot 2019-12-31 19 03 09
Screenshot 2019-12-31 19 02 54
Screenshot 2019-12-31 19 02 42

**With my commit, everything works fine, even when a BrokenPipeError is raised.** This can be shown by printing the exceptions that occur at the end of the run, and showing that all the animations still play. In the figure below, it printed [Errno 32] Broken pipe once, however, that error now does not damage the video.
image

Notes:
- If this error **never occurred** on some people's machines, this commit will **not change anything**, since it is simply catching the exception when an error is caused.
- Using `wait(1)` is mostly an arbitrary decision and seems to work well.
- If `sleep` is not called after `wait`, some animations do not show up on the screen.
- The changes in this commit **do not print** all the [Errno 32] Broken pipe exceptions that occur, which was done in the final figure.

This error also corresponds to the issue #845.

### DIFF
--- a/manimlib/scene/scene.py
+++ b/manimlib/scene/scene.py
@@ -2,6 +2,7 @@ import inspect
 import random
 import warnings
 import platform
+from time import sleep
 
 from tqdm import tqdm as ProgressDisplay
 import numpy as np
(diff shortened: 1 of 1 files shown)


manimlib/scene/scene.py
```diff
@@ -402,9 +403,15 @@ def handle_play_like_call(func):
...
-            self.file_writer.begin_animation(allow_write)
-            func(self, *args, **kwargs)
-            self.file_writer.end_animation(allow_write)
+            while True:
+                try:
+                    self.file_writer.begin_animation(allow_write)
+                    func(self, *args, **kwargs)
+                    self.file_writer.end_animation(allow_write)
+                    break
+                except:
```
PgBiel: Shouldn't you restrict this `except` temporarily to the error in question? This simply means that all errors will be ignored, which could be damaging depending on the task and the error.

PR #849, opened 2020-01-03
Align description to code

minus71:
The rotation happens in a clockwise direction, because the sign of the argument is negative.
---
I found it confusing and had to check the video to make sure the square was rotating clockwise.

So I thought to send a change.


TonyCrane:

PR #867, opened 2020-01-28
A simple Edit to readme File.

naveen521kk:
Changed from "python3 manim.py example_scenes.py SquareToCircle -pl" -> "py manim.py example_scenes.py SquareToCircle -pl" as python3 doesn't work on windows and only py works

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ python3 ./manim.py example_scenes.py SquareToCircle -pl
     git clone https://github.com/3b1b/manim.git
     cd manim
     pip3 install -r requirements.txt
-    python3 manim.py example_scenes.py SquareToCircle -pl
+    py manim.py example_scenes.py SquareToCircle -pl
     ```
 
 


zavden:
This is unnecessary, since `py` is nothing more than the name of the compiler, you can quietly change the name of the compiler from` py` to `python3`.

PgBiel:
In addition, `python3` is very much used on Unix-like systems (Linux, MacOS...), and not `py`, so this would make it valid for windows but not `python3`. (Perhaps the PR should suggest keeping both instead?)

naveen521kk:
IN typing python3 in cmd doesn't work. Only either py or python works.

zavden:
Yes, that is because the name of your compiler is py, but you can go to the folder where the compiler is located and change it. Also, the most common name for the compiler is python3 or python, it depends on your installation. So your PR is unnecessary, please close it.

Issue #884, opened 2020-02-13
How to choose a font with Text object?


qo4on:
This code does not change the font:
```
script = '''
AaBbCcDdEeFfGg
HhIiJjKkLlMmNn
OoPpQqRrSsTtUu
VvWwXxYyZz
'''

class WriteText(Scene):
  def construct(self):
    text = Text(script, font='/usr/share/fonts/truetype/open-sans/OpenSans-Regular.ttf')
    self.play(Write(text))
    self.wait(.1)
```
`!python3 -m manim example_scenes.py WriteText -r 480,854`
![image](https://user-images.githubusercontent.com/35609308/74466920-00a63e80-4eba-11ea-9bbb-cf11407616ab.png)

If I change this line the result is kind of similar to the target, but I can't choose `regular` and letters are distorted (look at the small `g` it has a stair on the top):
`text = Text(script, font="Open Sans")`
![image](https://user-images.githubusercontent.com/35609308/74467977-e5d4c980-4ebb-11ea-9e91-6ee2bc882828.png)

Elteoremadebeethoven:
See [this](https://github.com/3b1b/manim/pull/680#issuecomment-552732794)

qo4on:
@Elteoremadebeethoven Thank you.

I found that `Text` object can take:
`slant={ITALIC|NORMAL|OBLIQUE}`, `t2s={'world':ITALIC}`
`weight={BOLD|NORMAL}`, `t2w`
`size=5`
`lsh=1.5`
`color='#066E8B'`, `gradient=(BLUE, GREEN)`, `t2c={'[1:2]':'#e53125'}`, `t2g`
`fill_color`, `fill_opacity`, `stroke_color`, `stroke_width`

- What is the difference between `slant={ITALIC|OBLIQUE}` they look identical?
- Where can I find a list of all parameters that `Text` can take?
- As I see `font` parameter can't take path to the font thereby how can I choose a particular font style, for example `OpenSans-CondLightItalic.ttf`? I have several `Open Sans` fonts installed:
![image](https://user-images.githubusercontent.com/35609308/74513265-00df2200-4f2c-11ea-9ab2-ef782e191ddd.png)

Elteoremadebeethoven:
1. As far as I understand, ITALIC can have "ligarudas", that is, it is a separate source. And OBLIQUE simply transforms the normal source. So if your source does not have ITALIC support then it will look the same as OBLIQUE.

[...]

PR #680, opened 2019-08-12
New Text Mobject

xy-23:
# Motivation

Because the `TextMobject` is the subclass of `TexMobject`,

so it requires some latex-style behaviours in order to add some changes.

(e.g. other languages, font, slant and weight)

So I create `Text` mobject but you can still use `TextMobject`.

# Introduction

`Text` is a subclass of `SVGMobject`.

It uses `cairo` to generate text svgs.

It has a better UTF-8 support, multi-line support, and various options. 

It doesn't support mathematical formula(latex), 

you can use `TextMobject` or `TexMobject`.

You can create a text in manim like:
```
from manimlib.imports import *

class Demo(Scene):
    def construct(self):
        text = Text('Hello, world!')
        self.play(Write(text))
```

introduction

# Features

### text
Unlike `TextMobject`, `Text` only receive one `str`.

When there is only one line, the space(`' '`) in front or back will be ignored.

`'\t'` is not supported.

### color
`text = Text('Hello', color=BLUE)`

color

`color` defined in constants.py or a `str` like `'#FFFFFF'`, default is `WHITE`

### t2c (text2color)
`text = Text('Hello, world!', t2c={'world':BLUE})`

t2c

a `dict` like `{'text':color}` or Accurate mode

### gradient
`text = Text('Hello', gradient=(BLUE, GREEN))`

gradient

a `tuple` like `(color1, color2, color3, ...)`

### t2g (text2gradient)
`text = Text('Hello, world!', t2g={'world':(BLUE, GREEN)})`

t2g

a `dict` like `{'text':gradient}` or Accurate mode

### font
`text = Text('Hello', font='Source Han Sans')`

font

a `str`, the name of font like `'Source Han Sans'`, default is `CONFIG['font']`, defined in constants.py

Be sure you have the font installed!

### t2f (text2font)
`text = Text('Hello, world!', t2f={'world':'Forte'})`

t2f

a `dict` like `{'text':font}` or Accurate mode

Be sure you have the font installed!

### slant
`text = Text('Hello', slant=ITALIC)`

slant

`NORMAL` or `ITALIC`, default is `NORMAL`, defined in constants.py(a `str` actually)

### t2s (text2slant)
`text = Text('Hello, world!', t2s={'world':ITALIC})`

t2s

a `dict` like `{'text':slant}` or Accurate mode

### weight
`text = Text('Hello', weight=BOLD)`

weight

`NORMAL` or `BOLD`, default is `NORMAL`, defined in constants.py(a `str` actually)

[...]

PR #889, opened 2020-02-19
Fix Text wrong display

xy-23:
# Motivation
Relevant issues: https://github.com/3b1b/manim/issues/884#issue-564888972 & https://github.com/3b1b/manim/pull/680#issuecomment-552726606
I found that since #748 has merged, the `Text` Mobject was displayed unexpectedly.

# Behavior
Here is a simple demo reproduce the problem.
```

class Demo(Scene):
    def construct(self):
        text = Text('Hello, world!', font='Open Sans Condensed Light Italic')
        self.play(Write(text))
```
(Noticed that `stroke_width` was set to 1 by default.)
Before
Noticed there are some tiny "thorn".
Mark

# Solution
By setting `Text` Mobject's `stroke_width` to 0 default.
After
In order to avoid the similar problem, I also set some params to default.


xy-23:
@eulertour 
I have a new Pr that I want to push.
If there is nothing wrong, could you merged it?
Thanks : )

PR #839, opened 2019-12-25
Add implicit function graphing functionality

ddxtanx:
1. I made this pull request because I wanted, for my own personal work, to be able to graph implicit functions that were not parameterizable, and there was no functionality to do so in the current manim library. I coded it up and thought it would, hopefully, be useful enough for other people to use, and so I thought I'd share it in this pull request!
2. Here are various screenshots of various implicit functions that demonstrate this actually works, alone with some videos to demonstrate its runtime efficiency.

Graphing x^2 + y^2 = 4:
TestImplicitFuncs

Graphing y(x-y)^2 = 4x + 8
TestImplicitFuncs

Graphing y^2 = x^3 - x
TestImplicitFuncs

Animating y^2 = x^3 - x - b for b in [-1,1]
image

with runtime at around 4 minutes
image

Weirdly, the its/sec seemed to decrease over time for all of them, starting at around 2 its/sec and ending at 1.5 secs/it. I suspect something may be up with my generate_points algorithm, in that it bulks up memory somehow over time, but I haven't been able to find anything yet...

PR #911, opened 2020-03-12
Better discontinuity detecting and implicit function graphing

ddxtanx:
This is a merging of two pull requests to add better discontinuity detecting functionality and implicit function graphing capabilities in requests #832 and #839. I don't think it would serve much to repeat the rationale and testing again in another PR, so the relevant information is in the separate pull requests. If, however, it would be more conducive to reviewers to include such information here I would be happy to transfer it all over here.


Ammar34:
There is an issue with 1ed3c87. After saving changes to graph_scene.py, the plotted graph is not right as is evident from the attached files The first image is without changes to the file. The second one is with the changes in 1ed3c87 saved.
With Parametrarized function
without Parameterized function

PR #917, opened 2020-03-15
Update graph_scene.py

daniel-lara-ec:
This remove the decimal place by default and updates the label positions


TonyCrane:
Sorry, but `GraphScene` is no longer supported, its function has been moved to `CoordinateSystem`. So closed this.

PR #920, opened 2020-03-17
Set the axis tick label numbers to be the same color as the axis.

rpadams:

You should also include:

I changed the color of the GraphScene axis but could not change the color of the tick labels to match.  This is a small change to graph_scene.py to provide the correct information to NumberLine.

Before and after screenshots:
before
after


TonyCrane:
Sorry, but `GraphScene` is no longer supported, its function has been moved to` CoordinateSystem`. So closed this.

PR #950, opened 2020-03-30
Changed -1 to -1.0 for broader GPU suppot.

kolibril13:
Some GLSL compilers are more sensitive to types than others. With this fix, my Intel HD Graphics 5500 (Broadwell GT2) gives no error on this script anymore.

You should also include:


einarf:
Input from a moderngl maintainer:

There might be more of this around in the glsl code. I think where floats should be expected it's typically a good idea to use `1.` or `1.0` instead of `1` as it could lead to strange behaviour with some drivers and in this case actually a complier error.

This fix however makes Manim run at least!

zavden:
It works! thanks.

PR #961, opened 2020-04-06
Update sir.py

shagmwp:
Hi, I loved your Youtube video and thank you for your code. I updated the code to include Hospitalizations as well as Deceased.  This way you can see how scary a virus truly it. The common cold could have lots of infections, but no hospitalizations or deceased.  This way you can apply percentages of infected get hospitalized, and percentage of hospitalized pass away. With this you could add a "IF hospitals are over crowded death rate increase" condition (not added yet).

I also added an "at risk" attribute to model the percentage of population that are at higher risk to get hospitalized.

And lastly I added a healthy exposure simulation, where people not at risk can purposely expose themselves (think chicken pox parties) and then quarantine. You would be surprised.

I played with the models to increase infections time to 14 days, as well as increase time to symptomatic to 4 days, this would be more realistic  to Covid.

And lastly, I have one bug fix. Your last checking broke the multi city models.

You should also include:

### DIFF
--- a/from_3b1b/active/sir.py
+++ b/from_3b1b/active/sir.py
@@ -8,6 +8,8 @@ SICKLY_GREEN = "#9BBD37"
 COLOR_MAP = {
     "S": BLUE,
     "I": RED,
+    "H": ORANGE,
+    "D": YELLOW,
     "R": GREY_D,
 }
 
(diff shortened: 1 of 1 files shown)


Level-314:
Hi shagmwp - where did you find the fix_percent function? I do not see the beta_helpers source file in the repository.

3b1b:
This branch is being deleted, feel free to reopen the PR onto master.

PR #974, opened 2020-04-12
Performance improvement for most scenes

mikemag:
tl;dr: this is a significant performance improvement for many scenes. 1.7x - 2.6x improvement in animation it/s.

This is a small change to some of the hotest paths in rendering objects. The biggest win comes from not using np.allclose() to check if two points are close enough. In general, NumPy is awesome at operating on large arrays, but overkill for very tiny questions like this. Created a small function to determine if two points are close using the same algorithm, and limited it to 2D points since that's all we need in set_cairo_context_path().

A couple of other minor tweaks to reduce or eliminate other uses of NumPy in this path.

In general, it is better to avoid wrapping lists in np.array when a real NumPy array isn't actually needed.

Added a new file for performance test scenes, with a single scene from the end of a video I've been working on.

Data:

MacBook Pro (16-inch, 2019)
macOS Catalina 10.15.4
2.4 GHz 8-Core Intel Core i9
64 GB 2667 MHz DDR4
Python 3.7.3 (default, Mar  6 2020, 22:34:30)

Profiler: yappi under Pycharm.

Using the scene Perf1 from the included perf_scenes.py, averaged over 5 runs and rendered with:
manim.py perf_scenes.py Perf1 -pl --leave_progress_bars

Before:
Animation 0: FadeInTextMobject, etc.:               8.93it/s
Animation 1: ShowCreationParametricFunction, etc.: 84.66it/s

Profiler shows 48.5% of the run spent under Camera.set_cairo_context_path()

After
Animation 0: FadeInTextMobject, etc.:               23.45it/s  -- 2.63x improvement
Animation 1: ShowCreationParametricFunction, etc.: 149.62it/s  -- 1.77x improvement

Profiler shows 19.9% of the run spent under Camera.set_cairo_context_path()

Less improvement with production-quality renders, and percent improvement varies with scene of course. This appears to be a good win for every scene I'm working on though. I hope it will be for others, too.

NB: there are more perf improvements to be had, of course, but this is the best one I currently have.

You should also include:

### DIFF
--- a/manimlib/camera/camera.py
+++ b/manimlib/camera/camera.py
@@ -338,15 +338,15 @@ class Camera(object):
             return
 
         ctx.new_path()
-        subpaths = vmobject.get_subpaths_from_points(points)
+        subpaths = vmobject.gen_subpaths_from_points_2d(points)
         for subpath in subpaths:
-            quads = vmobject.get_cubic_bezier_tuples_from_points(subpath)
+            quads = vmobject.gen_cubic_bezier_tuples_from_points(subpath)
             ctx.new_sub_path()
             start = subpath[0]
             ctx.move_to(*start[:2])
             for p0, p1, p2, p3 in quads:
                 ctx.curve_to(*p1[:2], *p2[:2], *p3[:2])
-            if vmobject.consider_points_equals(subpath[0], subpath[-1]):
+            if vmobject.consider_points_equals_2d(subpath[0], subpath[-1]):
                 ctx.close_path()
         return self
 
@@ -549,7 +549,7 @@ class Camera(object):
     def transform_points_pre_display(self, mobject, points):
         # Subclasses (like ThreeDCamera) may want to
         # adjust points futher before they're shown
-        if np.any(np.isnan(points)) or np.any(points == np.inf):
+        if not np.all(np.isfinite(points)):
             # TODO, print some kind of warning about
             # mobject having invalid points?
             points = np.zeros((1, 3))
--- a/manimlib/mobject/types/vectorized_mobject.py
+++ b/manimlib/mobject/types/vectorized_mobject.py
@@ -595,35 +595,69 @@ class VMobject(Mobject):
             atol=self.tolerance_for_point_equality
         )
 
+    def consider_points_equals_2d(self, p0, p1):
+        """
+        Determine if two points are close enough to be considered equal.
+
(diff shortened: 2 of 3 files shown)


manimlib/camera/camera.py
```diff
@@ -338,15 +338,15 @@ def set_cairo_context_path(self, ctx, vmobject):
             return
 
         ctx.new_path()
-        subpaths = vmobject.get_subpaths_from_points(points)
+        subpaths = vmobject.gen_subpaths_from_points_2d(points)
```
eulertour, line 341: Wouldn't this give the incorrect answers for 3D scenes? I think you'll have to check whether or not this is being used with a 3D scene or not.

manimlib/mobject/types/vectorized_mobject.py
```diff
@@ -595,35 +595,69 @@ def consider_points_equals(self, p0, p1):
...
         )
 
+    def consider_points_equals_2d(self, p0, p1):
+        """
+        Determine if two points are close enough to be considered equal.
+
+        This uses the algorithm from np.isclose(), but expanded here for the
+        2D point case. NumPy is overkill for such a small question.
+        """
+        rtol = 1.e-5  # default from np.isclose()
```
eulertour, line 605: This should use `self.tolerance_for_point_equality`

manimlib/camera/camera.py
mikemag: We're in a 2D world at this point in `set_cairo_context_path()` and the Z value is just dead weight. Just above this call we have called `self.transform_points_pre_display()` to get the points, which the 3D camera uses to project objects onto 2D. Note that afterwards in this function the only use of `subpaths` is to eventually call `ctx.curve_to(*p1[:2], *p2[:2], *p3[:2])`, which only takes x,y coordinates, and the Z value is dropped from each point.

manimlib/mobject/types/vectorized_mobject.py
mikemag: `np.isclose()` takes two optional tolerances: atol and rtol. The original `consider_points_equals()` passes `self.tolerance_for_point_equality` for atol, and leaves rtol it's default value. I have done the same here; see lines 605 and 606 together.

`np.isclose()` def: https://github.com/numpy/numpy/blob/master/numpy/core/numeric.py#L2167

eulertour:
Alright, looks good!

PR #980, opened 2020-04-13
DOC: update Anaconda install instructions

mdhaber:
> The motivation for making this change (or link the relevant issues)

I use Anaconda on Windows, and I just tried to install manim, leaving open the possibility of hacking the source. The Anaconda installation instructions seemed a little out of date (maybe `environment.yml` has been updated since they were written?) so I thought I'd update them according to the procedure I needed to follow. I'll explain each change in a self-review.

> How you tested the new behavior (e.g. a minimal working example, before/after
screenshots, gifs, commands, etc.) 

Only by installing manim on my computer and confirming that the example runs.

Feel free to close if the instructions are fine as they are. I won't be hurt : )

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ python3 ./manim.py example_scenes.py SquareToCircle -pl
 
 4. [Install SoX](https://sourceforge.net/projects/sox/files/sox/).
 
-5. Install the remaining Python packages. Make sure that ``pycairo==1.17.1`` is changed to ``pycairo==1.18.0`` in requirements.txt.
+5. Install the remaining Python packages.
     ```sh
     git clone https://github.com/3b1b/manim.git
     cd manim
@@ -58,9 +58,13 @@ python3 ./manim.py example_scenes.py SquareToCircle -pl
 
 ## Anaconda Install
 
-* Install sox and latex as above. 
-* Create a conda environment using `conda env create -f environment.yml`
-* **WINDOWS ONLY** Install `pyreadline` via `pip install pyreadline`. 
+* Install SoX and LaTeX as above.
+* Clone this repository as above: `git clone https://github.com/3b1b/manim.git`.
+* Navigate into the repository as above: `cd manim`.
+* Create a conda environment using `conda env create -f environment.yml`.
+* Activate the environment: `conda activate manim`.
+* Add the repo's directory to the Python path: `conda develop .`.
+* Test your setup: `python -m manim example_scenes.py SquareToCircle -pl`.
 
 
 ### Using `virtualenv` and `virtualenvwrapper`


README.md
```diff
@@ -47,7 +47,7 @@ python3 ./manim.py example_scenes.py SquareToCircle -pl
 
 4. [Install SoX](https://sourceforge.net/projects/sox/files/sox/).
 
-5. Install the remaining Python packages. Make sure that ``pycairo==1.17.1`` is changed to ``pycairo==1.18.0`` in requirements.txt.
+5. Install the remaining Python packages.
```
mdhaber, line 50: `pycairo>=1.18.1; sys_platform == 'win32'` already in `requirements.txt`

README.md
```diff
@@ -58,9 +58,13 @@ python3 ./manim.py example_scenes.py SquareToCircle -pl
...
 
-* Install sox and latex as above. 
-* Create a conda environment using `conda env create -f environment.yml`
-* **WINDOWS ONLY** Install `pyreadline` via `pip install pyreadline`. 
+* Install SoX and LaTeX as above.
+* Clone this repository as above: `git clone https://github.com/3b1b/manim.git`.
+* Navigate into the repository as above: `cd manim`.
+* Create a conda environment using `conda env create -f environment.yml`.
+* Activate the environment: `conda activate manim`.
+* Add the repo's directory to the Python path: `conda develop .`.
```
mdhaber, line 66: This is useful if the user wants to write scripts outside of the `manim` root directory.

README.md
```diff
@@ -58,9 +58,13 @@ python3 ./manim.py example_scenes.py SquareToCircle -pl
...
 ## Anaconda Install
 
-* Install sox and latex as above. 
-* Create a conda environment using `conda env create -f environment.yml`
-* **WINDOWS ONLY** Install `pyreadline` via `pip install pyreadline`. 
+* Install SoX and LaTeX as above.
+* Clone this repository as above: `git clone https://github.com/3b1b/manim.git`.
+* Navigate into the repository as above: `cd manim`.
+* Create a conda environment using `conda env create -f environment.yml`.
+* Activate the environment: `conda activate manim`.
```
mdhaber, line 65: One could argue that self-contained instructions are not needed here, as anyone who is going to be messing with the source should know what steps are missing. But users other than developers might want to follow these instructions because Anaconda takes care of some of the dependencies.

README.md
```diff
@@ -58,9 +58,13 @@ python3 ./manim.py example_scenes.py SquareToCircle -pl
...
+* Add the repo's directory to the Python path: `conda develop .`.
+* Test your setup: `python -m manim example_scenes.py SquareToCircle -pl`.
```
mdhaber, line 67: I think it might be safer to just call `python` from an Anaconda environment. Perhaps it's something screwy in my setup, but for me, `python3` actually calls a version of python outside of the active environment. When in an environment, the environment's `python` takes priority over any other installed python; the same does not seem to be true of `python3`.

mdhaber:
OK, I'll go ahead and close : )

PR #996, opened 2020-04-20
Comment out SceneFromVideo

raghavg123:
`SceneFromVideo` apparently doesn't work and also adds the `opencv` dependency unnecessarily. This PR aims to exclude this class from `manimlib.imports` and make the opencv denedency optional.

You should also include:


xy-23:
Maybe you should also remove `opencv` in `environment.yml`, `requirments.txt` and `linux.rst`

TonyCrane:
This has been deleted in the new version, so I closed this.

Issue #941, opened 2020-03-27
Using braces


mvfs314:
I'd like to run the code:
`numeros4 = TexMobject(
			"\\mathbb{N}",
			"=",
			"\\{",
			"1",
			",",
			"2",
			",",
			"3",
			",",
			"\\ldots",
			"\\}"
		)`
cause I wish I could change the color of the braces separately. But I got this error:
`Exception: Latex error converting to dvi.`
I've already tried using the command `Brace()` or the ideas found at reddit, but it didn't work at all. Is there a way to make it possible?

Thanks!

Aathish04:
Have you installed LaTeX? You also need quite a few LaTeX packages with it as well.
Please post the full traceback and we may be able to help you better.

mvfs314:
I certainly have. This is the basic LaTeX code, I just wanna write the braces separately. 

Here the traceback:
`Traceback (most recent call last):
  File "C:\Manim\manim-master\manimlib\extract_scene.py", line 155, in main
    scene = SceneClass(**scene_kwargs)
  File "C:\Manim\manim-master\manimlib\scene\scene.py", line 53, in __init__
    self.construct()
  File "1.py", line 170, in construct
    self.numeros_naturais()
  File "1.py", line 241, in numeros_naturais
    "\\}"
  File "C:\Manim\manim-master\manimlib\mobject\svg\tex_mobject.py", line 153, in __init__
    self.break_up_by_substrings()
  File "C:\Manim\manim-master\manimlib\mobject\svg\tex_mobject.py", line 184, in break_up_by_substrings
    sub_tex_mob = SingleStringTexMobject(tex_string, **config)
  File "C:\Manim\manim-master\manimlib\mobject\svg\tex_mobject.py", line 44, in __init__
    self.template_tex_file_body
  File "C:\Manim\manim-master\manimlib\utils\tex_file_writing.py", line 19, in tex_to_svg_file
    dvi_file = tex_to_dvi(tex_file)
  File "C:\Manim\manim-master\manimlib\utils\tex_file_writing.py", line 67, in tex_to_dvi
    "See log output above or the log file: %s" % log_file)
Exception: Latex error converting to dvi. See log output above or the log file: ./media\Tex\0447ed74899f41fb.log`

Thanks!

Aathish04:
Hmm, the traceback points to a log file in `./media\Tex\0447ed74899f41fb.log`. Could you please post the contents of that as well?

[...]

PR #1013, opened 2020-04-29
CodeMobject to display code with color highlighted

NavpreetDevpuri:
you can use it as follow.
format code from https://codebeautify.org/cpp-formatter-beautifier with select intent = 2 means tab spacing = 2
save html file from https://tohtml.com/cpp/ select style = 'black'
```python
from manimlib.mobject.code_mobject import *
code = CodeMobject("Helloword.html",
                           coordinates=[-5,3],
                           run_time=3,
                           scale_factor=0.3,
                           line_spacing=0.2,
                           tab_spacing=0.6,
                           font='Monospac821 BT')
        #self.draw_code_all_lines_at_a_time(code)
        code.scale(1)
        self.play(Write(code),run_time=1)

    def draw_code_all_lines_at_a_time(self, code):
        self.play(*[Write(code[i]) for i in range(code.__len__())],run_time=code.run_time)
```
coordinate point is LEFT+UP corner. 
'code' is a 2d array of characters.
for example to access first 5 characters from line 2 use `code[2][0:5]`
```python
code[2][0:5].set_color(RED)
```

Output will be looks like 
code

* It solves the following problems
1) display code with color highlighted
2) manage to print single LaTeX character '{' or '}' because of using Text() instead TextMobject() from https://github.com/3b1b/manim/issues/941#issuecomment-621615699


tshrpl:
wow! this will be really usefull if the pull request is accepted ...

safinsingh:
It would be nice if you could change the font specifically for this Mobject to make it look more like code!

NavpreetDevpuri:
> It would be nice if you could change the font specifically for this Mobject to make it look more like code!

Updated !

NavpreetDevpuri:
https://github.com/3b1b/manim/issues/1017

it have that issue.

NavpreetDevpuri:
Reopening it with new branch

PR #1014, opened 2020-04-29
fixed bug for previewing gif files

tshrpl:
# motivation
when `-i` is used with `-p`, the manim script ignores the gif file and opens `.mp4` with the same name
this pr fixes the bug

# test
[video](https://www.youtube.com/watch?v=K-OSIek7xkY)


tshrpl:
oops, `-i` is redundant now ... see 61bb49

PR #1018, opened 2020-05-01
Code() in file Code_mobject.py to display code with color highlighted works correctly with updates in text_mobject.py and svg_mobject. 

NavpreetDevpuri:
you can use it as follow.
if you put file into "assets/codes" folder then you don't need to specify full file path, just name of file is enough otherwise specify full file path.
`if generate_html_file==True` it create a html file with color  highlighted and save it to "assets/codes/generated_html_files/" 
```python
from manimlib.imports import *
class codeExample(Scene):
    def construct(self):
        heading = TextMobject("\"Hello, World\" Program", stroke_width=0).scale(1.3)
        heading.to_edge(UP)
        helloworldc = Code("helloworldc.c",
                           run_time=1,
                           scale_factor=0.6,
                           line_spacing=0.2,
                           tab_spacing=0.6,
                           insert_line_no=False,
                           style=code_styles_list[4],
                           background="rectangle",
                           language=code_languages_list["c"],
                           generate_html_file=True
                           )
        helloworldcpp = Code("helloworldcpp.cpp",
                             run_time=1,
                             scale_factor=0.6,
                             line_spacing=0.2,
                             tab_spacing=0.6,
                             margin=0.3,
                             insert_line_no=True,
                             line_no_from=8,
                             line_no_buff=0.4,
                             style=code_styles_list[9],
                             background="window",
                             corner_radius=0.2,
                             language=code_languages_list["cpp"],
                             generate_html_file=True
                             )
        helloworldc.move_to(np.array([-3.6, 0, 0]))
        helloworldcpp.move_to(np.array([3.1, 0, 0]))
        self.play(Write(heading), run_time=0.5)
        self.play(Write(helloworldc), run_time=1.3)
        self.draw_code_all_lines_at_a_time(helloworldcpp)
        self.wait()

    def draw_code_all_lines_at_a_time(self, Code):
        self.play(Write(Code.background_mobject), run_time=0.3)
        self.play(Write(Code.line_numbers), run_time=0.3)
        self.play(*[Write(Code.code[i]) for i in range(Code.code.__len__())],
                  run_time=Code.run_time)

```
Output will be looks like 
ezgif-6-3eb6f393e805

```Python
'''
coordinate point is LEFT+UP corner

Code is VGroup() with three things
Code[0] is Code.background_mobject
    which can be a 
        Rectangle() if background == "rectangle" 
        VGroup() of Rectangle() and Dot() for three buttons
Code[1] is Code.line_numbers
[...]

PR #1031, opened 2020-05-05
Fix Text Class's bugs in shaders branch (the same as #1030)

Tony031218:
Fixed Three bugs of Text Class.
This pull request is the same as #1030
Without this, Text Class wouldn't work in the 'shaders' branch.

### DIFF
--- a/manimlib/mobject/svg/text_mobject.py
+++ b/manimlib/mobject/svg/text_mobject.py
@@ -73,7 +73,8 @@ class Text(SVGMobject):
             self.set_color_by_t2g()
 
         # anti-aliasing
-        self.scale(TEXT_MOB_SCALE_FACTOR)
+        if self.height is None:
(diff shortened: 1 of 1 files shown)


NavpreetDevpuri:
i also changed text_mobject.py in https://github.com/3b1b/manim/pull/1018 
i also update file to newer version but in file changes it shows that i added also lines that you added, but its should show only my changes. does it create any problem or its okay ?

Tony031218:
> i also changed text_mobject.py in #1018
> i also update file to newer version but in file changes it shows that i also added lines that you added, but its should show only my changes. does it create any problem or its okay ?

This pull request is of shaders branch not master. If you want update text_mobject.py you can just pull the newer version from master, because my pull request #1030 has already been merged.

NavpreetDevpuri:
thanks 
understood!

PR #1032, opened 2020-05-05
add raw frame operation scene

niedong:
Video consists of frames, and in most cases people don't have to know or realize it. manim library helps us handle it so we don't have to care about lower level details. However, there are a few cases where we want to make videos by continuously adding massive amount of mobjects on the screen(e.g. They are created by some mathematical laws, often in large quantities and in small sizes). Simply using 'self.add(...)' method would requires way more time to produce each frame because they are all added into 'mobjects' attribute. Not only does 'self.add(...)' take time, but also we repaint them on the screen each time we write an animation. We can help user avoid this by providing raw frame operation interface. Users don't have to waste their time on 'self.add(...)' call, and each time they add objects on the frame, manim library doesn't have to repaint them over and over again.

I actually met this problem when I want to make a video about solitaire waterfall(classic animation on Windows solitaire game).  I was using 'self.add(...)' method and realized the problem: I spent several days and it couldn't finish the animation. So I wrote this 'raw_frame_scene.py', manipulating each frame. And it just took me a few minutes to finish the animation.

I uploaded the video on 'https://www.bilibili.com/video/BV12a4y1x78t', and the source code is on 'https://github.com/manim-kindergarten/manim_sandbox/tree/master/videos/Solitaire'. It works perfectly. I really wish manim library could also have this special scene because it is really useful on particular cases.

### DIFF
--- a/example_scenes.py
+++ b/example_scenes.py
@@ -5,7 +5,7 @@ from manimlib.imports import *
 # To watch one of these scenes, run the following:
 # python -m manim example_scenes.py SquareToCircle -pl
 #
-# Use the flat -l for a faster rendering at a lower
+# Use the flag -l for a faster rendering at a lower
 # quality.
 # Use -s to skip to the end and just save the final frame
 # Use the -p to have the animation (or image, if -s was


Tony031218:
This is a really a good way to improve the efficiency of making frame-by-frame animation.
But I think it will be much better if you add a feature to choose whether clear the buffer every frame automatically, because sometimes we don't want to still leave the buffer on the scene like Solitaire.
By the way, you'd better add `from manimlib.scene.raw_frame_scene imports *` to `imports.py`.

kevkevinpal:
I'm not sure if this is related but if the issue could we not add mobjects onto a cache and then check to see if they match the value in the cache and if it doesn't then rewrite to the screen?

niedong:
> I'm not sure if this is related but if the issue could we not add mobjects onto a cache and then check to see if they match the value in the cache and if it doesn't then rewrite to the screen?

"rewrite to the screen" is not something you can control. To support updater every mobject has to be repainted. You can look at how "self.clear" method is implemented.

niedong:
Try the following simple code. It is exactly what I'm talking about:
```python
class TestScene(Scene):
    def construct(self):
        self.add(*[
            Rectangle()
            for i in range(100000)
        ])
        for i in range(10000):
            self.add(Dot())
            print("Using 'self.add' method...{}".format(i + 1), end="\r")
```
This code aims to show that when there are already 100000 mobjects in the attribute, the following 'self.add(...)' call could take much more time than you expected. By implementing raw frame operation scene, this problem can be avoided perfectly.

niedong:
> Try the following simple code. It is exactly what I'm talking about:
> 
> ```python
> class TestScene(Scene):
>     def construct(self):
>         self.add(*[
>             Rectangle()
>             for i in range(100000)
>         ])
>         for i in range(10000):
>             self.add(Dot())
>             print("Using 'self.add' method...{}".format(i + 1), end="\r")
> ```
> 
> This code aims to show that when there are already 100000 mobjects in the attribute, the following 'self.add(...)' call could take much more time than you expected. By implementing raw frame operation scene, this problem can be avoided perfectly.

Where there's a need to use 'self.add(...)' call like this, expectially in large quantities, replace it with 'self.capture(...)'(using RawFrameScene), the latter will perform 10 (or more) times better.

kevkevinpal:
> > Try the following simple code. It is exactly what I'm talking about:
> > 
> > ```python
> > class TestScene(Scene):
> >     def construct(self):
> >         self.add(*[
> >             Rectangle()
> >             for i in range(100000)
> >         ])
> >         for i in range(10000):
> >             self.add(Dot())
> >             print("Using 'self.add' method...{}".format(i + 1), end="\r")
> > ```
> > 
> > This code aims to show that when there are already 100000 mobjects in the attribute, the following 'self.add(...)' call could take much more time than you expected. By implementing raw frame operation scene, this problem can be avoided perfectly.
> 
> Where there's a need to use 'self.add(...)' call like this, especially in large quantities, replace it with 'self.capture(...)'(using RawFrameScene), the latter will perform 10 (or more) times better.

Sorry if I didn't look too deeply into this issue but if this is doing essentially the same thing as the .add why don't we implement it into the .add so a use wouldn't need to know about the trade offs and rather be able to just use .add and have it work efficiently. I'd just rather have less files to think about and have the library be smaller. Again sorry if I'm not fully grasping what this RawFrameScene is getting at

niedong:
> > > Try the following simple code. It is exactly what I'm talking about:
> > > 
> > > ```python
> > > class TestScene(Scene):
> > >     def construct(self):
> > >         self.add(*[
> > >             Rectangle()
> > >             for i in range(100000)
> > >         ])
> > >         for i in range(10000):
> > >             self.add(Dot())
> > >             print("Using 'self.add' method...{}".format(i + 1), end="\r")
> > > ```
> > > 
> > > This code aims to show that when there are already 100000 mobjects in the attribute, the following 'self.add(...)' call could take much more time than you expected. By implementing raw frame operation scene, this problem can be avoided perfectly.
> > 
> > Where there's a need to use 'self.add(...)' call like this, especially in large quantities, replace it with 'self.capture(...)'(using RawFrameScene), the latter will perform 10 (or more) times better.
> 
> Sorry if I didn't look too deeply into this issue but if this is doing essentially the same thing as the .add why don't we implement it into the .add so a use wouldn't need to know about the trade offs and rather be able to just use .add and have it work efficiently. I'd just rather have less files to think about and have the library be smaller. Again sorry if I'm not fully grasping what this RawFrameScene is getting at

What rawframescene does is it keeps updating current frame, and write it. Thats' it. 'self.add()' has its own unique usage, depending on mechanism of manim library, and the two have little connections frankly. What I'm talking about is that there's actually a weakness in manim library. Fix it would make it better and more powerful. As you can see, this code is clear and small-sized.

kevkevinpal:
> > > > Try the following simple code. It is exactly what I'm talking about:
> > > > 
> > > > ```python
> > > > class TestScene(Scene):
> > > >     def construct(self):
> > > >         self.add(*[
> > > >             Rectangle()
> > > >             for i in range(100000)
> > > >         ])
> > > >         for i in range(10000):
> > > >             self.add(Dot())
> > > >             print("Using 'self.add' method...{}".format(i + 1), end="\r")
> > > > ```
> > > > 
> > > > This code aims to show that when there are already 100000 mobjects in the attribute, the following 'self.add(...)' call could take much more time than you expected. By implementing raw frame operation scene, this problem can be avoided perfectly.
> > > 
> > > Where there's a need to use 'self.add(...)' call like this, especially in large quantities, replace it with 'self.capture(...)'(using RawFrameScene), the latter will perform 10 (or more) times better.
> > 
> > Sorry if I didn't look too deeply into this issue but if this is doing essentially the same thing as the .add why don't we implement it into the .add so a use wouldn't need to know about the trade offs and rather be able to just use .add and have it work efficiently. I'd just rather have less files to think about and have the library be smaller. Again sorry if I'm not fully grasping what this RawFrameScene is getting at
> 
> What rawframescene does is it keeps updating current frame, and write it. Thats' it. 'self.add()' has its own unique usage, depending on mechanism of manim library, and the two have little connections frankly. What I'm talking about is that there's actually a weakness in manim library. Fix it would make it better and more powerful. As you can see, this code is clear and small-sized. 

So is self.capture writing to exactly one frame multiple times? Where as .add is adding mobjects to the scene but doesn't really take into account how many objects are being added so it takes up multiple frames which can cause issues if adding large quantities of mobjects? If so then that makes complete sense and is awesome.

niedong:
> > > > > Try the following simple code. It is exactly what I'm talking about:
> > > > > ```python
> > > > > class TestScene(Scene):
> > > > >     def construct(self):
> > > > >         self.add(*[
> > > > >             Rectangle()
> > > > >             for i in range(100000)
> > > > >         ])
> > > > >         for i in range(10000):
> > > > >             self.add(Dot())
> > > > >             print("Using 'self.add' method...{}".format(i + 1), end="\r")
> > > > > ```
> > > > > 
> > > > > 
> > > > > This code aims to show that when there are already 100000 mobjects in the attribute, the following 'self.add(...)' call could take much more time than you expected. By implementing raw frame operation scene, this problem can be avoided perfectly.
> > > > 
> > > > 
> > > > Where there's a need to use 'self.add(...)' call like this, especially in large quantities, replace it with 'self.capture(...)'(using RawFrameScene), the latter will perform 10 (or more) times better.
> > > 
> > > 
> > > Sorry if I didn't look too deeply into this issue but if this is doing essentially the same thing as the .add why don't we implement it into the .add so a use wouldn't need to know about the trade offs and rather be able to just use .add and have it work efficiently. I'd just rather have less files to think about and have the library be smaller. Again sorry if I'm not fully grasping what this RawFrameScene is getting at
> > 
> > 
> > What rawframescene does is it keeps updating current frame, and write it. Thats' it. 'self.add()' has its own unique usage, depending on mechanism of manim library, and the two have little connections frankly. What I'm talking about is that there's actually a weakness in manim library. Fix it would make it better and more powerful. As you can see, this code is clear and small-sized.
> 
> So is self.capture writing to exactly one frame multiple times? Where as .add is adding mobjects to the scene but doesn't really take into account how many objects are being added so it takes up multiple frames which can cause issues if adding large quantities of mobjects? If so then that makes complete sense and is awesome. (Maybe we can throw a warning in the .add method if we notice an array being passed with a large quantity of mobjects and point it to this RawFrameScene so a user doesn't have to dig through docs or find it themselves)

Yes, exactly one frame(but updated). So it by-pass the 'self.add()' call, which can take way more time if there's already large quantities of mobjects in the attribute. At this point we can go back to the essence of every video, and that is frame. No matter how many mobjects we want to add on the screen, we simply update them on the current frame, and it takes fixed and less time.

huguesdevimeux:
> I'm not sure if this is related but if the issue could we not add mobjects onto a cache and then check to see if they match the value in the cache and if it doesn't then rewrite to the screen?

Sorry for coming late to the party, but this is now implemented on ManimCommunity. You should take a look :)

Issue #1017, opened 2020-05-01
Text(" ") don't move. Because of that Text("a   b") shows wrong transform animation.


NavpreetDevpuri:
```python 
class test(Scene):
    def construct(self):
        text = Text("     ")
        text.to_corner(DOWN+LEFT)
        rect = SurroundingRectangle(text)
        self.add(text,rect)
```
## Output 
![test](https://user-images.githubusercontent.com/30471072/80776058-f7ef0b80-8b7e-11ea-9827-e20dbea176bc.png)

```python
class test(Scene):
    def construct(self):
        text = Text("a     b")
        text1 = Text("123")
        text.to_corner(DOWN+LEFT)
        text1.to_edge(RIGHT+DOWN)
        rect = SurroundingRectangle(text)
        self.add(text,rect)
        self.play(Transform(text,text1))
        self.wait()
```
## Output
![ezgif-6-9891bd47c1c2](https://user-images.githubusercontent.com/30471072/80776516-320cdd00-8b80-11ea-8f1e-50fddab4995c.gif)

NavpreetDevpuri:
## i check text's svg file it looks like as follow
```html
<path style="stroke:none;" d="M 4.042969 -0.640625....ect"/>
<!-- HERE IS THE PROBLEM A PATH WITH NO VALUE FOR 'd' ATTRIBUTE -->
<path style="stroke:none;" d=""/>
<path style="stroke:none;" d="M 1.46875 0 L 0.65625....etc"/>
```

NavpreetDevpuri:
In file svg_mobject.py
```python
        elif element.tagName == 'path':
            result.append(self.path_string_to_mobject(
                element.getAttribute('d')
            ))
```

NavpreetDevpuri:
https://github.com/3b1b/manim/pull/1018

Solved.

NavpreetDevpuri:
Now it works fine.
![ezgif-6-88395b75ecac](https://user-images.githubusercontent.com/30471072/80785832-35fc2780-8b9f-11ea-972e-34f1a159bed0.gif)

kevkevinpal:
is there any action still needing to be done if not can we close this issue? @NavpreetDevpuri

NavpreetDevpuri:
change file manimlib/mobject/svg/svg_mobject.py from line 89
```python
            elif element.tagName == 'path':
            # From this 
            '''
            result.append(self.path_string_to_mobject(
                element.getAttribute('d')
            ))
            '''
[...]

PR #1036, opened 2020-05-07
Code() in file Code_mobject.py to display code with color highlighted added Paragraph() and "exact_spaces" parameter to Text() 

NavpreetDevpuri:
## It solves the following problems
1) display code with color highlighted
2) manage to print single '{' or '}'  because of using Text() instead of TextMobject() from https://github.com/3b1b/manim/issues/941#issuecomment-621615699
3) Solved Text( ) transform animation for " " space character from https://github.com/3b1b/manim/issues/1017
4) Text() with tabs, indexes not works correctly https://github.com/3b1b/manim/issues/1060
5) Text() don't shows spaces in front of the text and at the end of the text https://github.com/3b1b/manim/issues/1061

you can use it as follow.
if you put file into "assets/codes" folder then you don't need to specify full file path, just name of file is enough otherwise specify full file path.
`if generate_html_file==True` it create a html file with color  highlighted and save it to "assets/codes/generated_html_files/" 
```python
class codeExample(Scene):
    def construct(self):
        heading = TextMobject("\"Hello, World\" Program", stroke_width=0).scale(1.3)
        heading.to_edge(UP)
        helloworldc = Code("helloworldc.c",
                           run_time=1,
                           scale_factor=0.6,
                           line_spacing=0.2,
                           tab_spacing=0.6,
                           insert_line_no=False,
                           style=code_styles_list[4],
                           background="rectangle",
                           language=code_languages_list["c"],
                           generate_html_file=True
                           )
        helloworldcpp = Code("helloworldcpp.cpp",
                             run_time=1,
                             scale_factor=0.6,
                             line_spacing=0.2,
                             tab_spacing=0.6,
                             margin=0.3,
                             insert_line_no=True,
                             line_no_from=8,
                             line_no_buff=0.4,
                             style=code_styles_list[9],
                             background="window",
                             corner_radius=0.2,
                             language=code_languages_list["cpp"],
                             generate_html_file=True
                             )
        helloworldc.move_to(np.array([-3.6, 0, 0]))
        helloworldcpp.move_to(np.array([3.1, 0, 0]))
        self.play(Write(heading), run_time=0.5)
        self.play(Write(helloworldc), run_time=1.3)
        self.draw_code_all_lines_at_a_time(helloworldcpp)
        self.wait()

    def draw_code_all_lines_at_a_time(self, Code):
        self.play(Write(Code.background_mobject), run_time=0.3)
        self.play(Write(Code.line_numbers), run_time=0.3)
        self.play(*[Write(Code.code[i]) for i in range(Code.code.__len__())],
                  run_time=Code.run_time)

```
Output will be looks like 
ezgif-6-3eb6f393e805

```Python
'''
1) Code is VGroup() with three things
    1.1) Code[0] is Code.background_mobject
        which can be a 
            1.1.1) Rectangle() if background == "rectangle" 
            1.1.2) VGroup() of Rectangle() and Dot() for three buttons if background == "window" 
    1.2) Code[1] is Code.line_numbers Which is a Paragraph() object, this mean you can use 
                Code.line_numbers[0] or Code[1][0] to access first line number 
    1.3) Code[2] is Code.code
        1.3.1) Which is a Paragraph() with color highlighted, this mean you can use 
            Code.code[1] or Code[2][1] 
                line number 1
            Code.code[1][0] or Code.code[1][0] 
                first character of line number 1
            Code.code[1][0:5] or Code.code[1][0:5] 
                first five characters of line number 1
'''
```
Created Paragraph()
```python
class Test3(Scene):
    def construct(self):
        t = Paragraph('this is a awesome', 'paragraph', 'With \nNewlines', '\tWith Tabs', '  With Spaces', 'With Alignments',
                 'center', "left", "right", line_spacing=0.1, alignment="left", t2c={"Tabs": RED})
        t.set_alignment("center", 7)
        t.set_alignment("left", 8)
        t.set_alignment("right", 9)
        #t.set_all_lines_alignment("center")
        t[0][5].set_color(GREEN)
        t[0][6].set_color(GREEN)
        t[1][0:4].set_color(YELLOW)
        rect = SurroundingRectangle(t)
        self.add(t,rect)
        self.wait()
```
Output 
Test6

```python
class Test7(Scene):
    def construct(self):
        t = Paragraph('this is a awesome', 'paragraph', 'With \nNewlines', '\tWith Tabs', '  With Spaces',
                  'With Alignments',
                  'center', "left", "right", line_spacing=0.1, alignment="left", t2c={"Tabs": RED})
        #t.set_alignment("center", 7)
        self.add(t)
        self.play(t.shift, RIGHT)
        rect = SurroundingRectangle(t)
        self.add(rect)
        self.play(t.set_alignment, "right", 9)
        self.play(t.set_all_lines_alignment, "center")
        self.wait()
```
Output 
ezgif-1-ec983ee94268

Fixed https://github.com/3b1b/manim/issues/1060 and https://github.com/3b1b/manim/issues/1061
```python
class Test4(Scene):
    def construct(self):
        t2 = Text(" a\ty  ").scale(3)
        t2[3].set_color(GREEN)
        r = SurroundingRectangle(t2)
        self.add(t2,r)
        self.wait()
``` 
Output
Test4


Tony031218:
In my opinion, there is no need to turn `Text` into `SingleStringTextMobject`, that will cause some problems when users using Text from old version to new version.

It is better to create a new class `Texts` to archive the same effect of `Text` now which archives more functions. Do you undersand me?
```text
SingleStringTextMobject -> Text
Text -> Texts
```

NavpreetDevpuri:
> SingleStringTextMobject -> Text
> Text -> Texts
Great !

xy-23:
Well, I have implemented the multiple text parameters in `Text` just now...
And I think it's a better way compared to your current solution.
So, I think we don't need `Texts` nor `SingleStringTextMobject`.
But I need to fix the space problem first until make another PR to update it.

Tony031218:
> Well, I have implemented the multiple text parameters in `Text` just now...
> And I think it's a better way compared to your current solution.
> So, I think we don't need `Texts` nor `SingleStringTextMobject`.
> But I need to fix the space problem first until make another PR to update it.

Yes, the space problem is a **bug** of Text which must to be fixed, but code(or Texts) is a new **feature**.

NavpreetDevpuri:
> Yes, the space problem is a **bug** of Text which must to be fixed, but code(or Texts) is a new **feature**.

spaces and tabs problem is solved by just giving those characters a physical height and width by putting appropriate invisible Rectangle so that those characters are also select able for example by using SurroundingRectangle()

xy-23:
The reason not giving those space characters a physical height and width is that
they don't have a static size at all. Their size are dynamic.
image
So, how could you even decide the size of those characters which are in the front or back of letters.

xy-23:
And from your reply, it seems that without displaying the space in the front or back of text, your Code() can behave correctly.

NavpreetDevpuri:
I calculated the width of space according to font a

> The reason not giving those space characters a physical height and width is that
> they don't have a static size at all. Their size are dynamic.
> image
> So, how could you even decide the size of those characters which are in the front or back of letters.

It calculate the width of space character according to font.

xy-23:
But calculating the width of space according to font a is not rigorous.

NavpreetDevpuri:
> But calculating the width of space according to font a is not rigorous.

check new method get_space_width() at line 233 in text_mobject.py to calculate space_width which calculates the exact width of space according to font type or language type
```python
class Test4(Scene):
    def construct(self):
        t = Text("a c").scale(3.6)
        r = SurroundingRectangle(t[1],stroke_width=1,buff=0)
        self.add(t, r)
        self.wait()
```
output 
Test4

xy-23:
Seems my PR is merged. But I have to say something.

I think the main disagreement between us is that whether we should display those space characters in the front or back of letters. I think that I should emphasize here that I am saying the situation where there is only single line of letters, more specifically，this picture's situation.
image
I contradict this because, like I said, the size of those characters are dynamic. And the text layout engine that currently Text uses will ignore them directly. I knows little about the layout process under the engine, so I decided to do nothing, not displaying the empty space is the current solution.

And you think we should display those empty space. I understand. And from the user's perspective, I definitely agree that you said `it should be display exactly same as it is written by user`. But that is hard to achieve, because the engine we are using can't provide this information at all! And your solution is not a good choice because it adds potential problems to `Text()`which we are not sure.

So, I'm trying to persuade you to give up fixing the space characters problem in svg_mobject.py and text_mobject.py. Your `Code()` have already good enough. And thanks you for contributing manim.

I think you probably need this. I noticed the way you are dealing with multiple-lines text, you separate them by split('\n'), and sent them to `Text()` one by one, that's why you need to display those empty characters (the situation above), am I right? Well, there is a better way to do this, you can send the multiple-lines text directly to `Text()` and separate (or group) them via Text.submobjects, this means you can have the "space size information". You can implement this as following:
```py
    def set_group(self):
        new_submobjects = []
        start = 0
        for text in self.texts:
            submobjects = self.submobjects[start:start+len(text)]
            new_submobjects.append(VGroup(*submobjects))
            start += len(text)
        self.submobjects = new_submobjects
```
Yes, this would lose the ability to align manually. But it's not very important. And I should also notice you that this would let `t2c` failed to work. It's just an example.

The reason why I don't want you to change the `Text()` is that I don't want `Text()` to be a another `TextMobject()` which is hard to use (in some way). And in order to keep its compatibility as well. What's more, I'm planning to switch to another text layout engine, which can not only display those empty space characters correctly, but also provide more features like alignment, emoji-support and so on.

As for your `Code()`, I still think that it would be much better if you can separate the features that `Code()` provides. I think `Code()` is just a colored `Text()`, and line numbers can be another Mobject such as calling it `LineNum()`,  and the background, should be handle by something like `Window()`， rather than by `Code()` itself. But that's just my suggestions. It depends on you.

Tony031218:
I agree with xy-23, I don’t think it is a good idea to change too many codes of `Text()`.

But the feature which displays space characters in front of text may be very useful to your `Code()`. 

If ‘the feature’ is very necessary to your `Code()`, I think it’s better to create a new subclass of `Text()`.
You only have to rewrite some methods to archive your goals. In this way, `text_mobject.py` don’t need to be changed, and you can create `Code()` as well.

By the way, I think `code_mobject.py` is better in folder `mobject/svg/` (the same as `text_mobject.py` ) than in folder `mobject/`

NavpreetDevpuri:
Added a new parameter "exact_spaces" to Text()

NavpreetDevpuri:
> And you think we should display those empty space. I understand. And from the user's perspective, I definitely agree that you said `it should be display exactly same as it is written by user`. But that is hard to achieve, because the engine we are using can't provide this information at all! And your solution is not a good choice because it adds potential problems to Text()which we are not sure.

i did that i just calculated space_width as `the engine` is drawing those space character and i just calculate that width. This will work fine even with different languages or fonts.

NavpreetDevpuri:
@eulertour any suggestions ?

NavpreetDevpuri:
> As for your `Code()`, I still think that it would be much better if you can separate the features that `Code()` provides. I think `Code()` is just a colored `Text()`, and line numbers can be another Mobject such as calling it `LineNum()`, and the background, should be handle by something like `Window()`， rather than by `Code()` itself. But that's just my suggestions. It depends on you.

by creating new classes ?

Tony031218:
@NavpreetDevpuri you forgot add `pygments` to `environment.yml` which used for anaconda. See #1136

NavpreetDevpuri:
@Tony031218 
check [this](https://github.com/NavpreetDevpuri/manim/blob/master/environment.yml)
is this alright ?

NavpreetDevpuri:
btw @efroim102 should merge https://github.com/3b1b/manim/pull/1071 as soon as possible.
because it fixes very big problems.

PR #1039, opened 2020-05-10
Corrected a bug in stretch_to_fit_depth func

peninha:
The "stretch_to_fit_depth" function is calling "rescale_to_fit" with the wrong dimension. Just corrected that.

You should also include:


Tony031218:
#1038 is merged

PR #1052, opened 2020-05-11
Dashed pattern using pycairo.

andre-caldas:
I have created a dashed stroke pattern.
By now, dashed patterns are made by splitting a curve into many small pieces.
This patch uses pycairos stroke configuration.
DashedLine() should be patched, too.

Using it is very simple:
self.play(
    ShowCreation(Circle(radius=1.0, dash_pattern=DASHED)),
    ShowCreation(Circle(radius=2.0, dash_pattern=[.4, 0.05, 0.1, 0.05])),
)

Reference:
https://www.cairographics.org/samples/dash/

[dashed.py.txt](https://github.com/3b1b/manim/files/4612328/dashed.py.txt)
Screenshot 2020-05-11 18-59-41

### DIFF
--- a/manimlib/camera/camera.py
+++ b/manimlib/camera/camera.py
@@ -396,7 +396,12 @@ class Camera(object):
             # as you zoom in on them.
             (self.get_frame_width() / FRAME_WIDTH)
         )
+        dash_pattern = vmobject.get_stroke_dash_pattern()
+        ctx.save()
+        if None != dash_pattern:
+            ctx.set_dash(dash_pattern)
         ctx.stroke_preserve()
+        ctx.restore()
         return self
 
     def get_stroke_rgbas(self, vmobject, background=False):
--- a/manimlib/constants.py
+++ b/manimlib/constants.py
@@ -195,6 +195,9 @@ DEGREES = TAU / 360
 
 FFMPEG_BIN = "ffmpeg"
 
+# Dashes
+DASHED = [.1, .05]
+
 # Colors
 COLOR_MAP = {
     "DARK_BLUE": "#236B8E",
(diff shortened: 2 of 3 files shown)


Tony031218:
Why not use `DashedVMobject()`.

andre-caldas:
> Why not use DashedVMobject().

1. The DashedVMobject() does not allow one to setup a dash pattern.
The patterns in the screenshot I have attached to the pull request
cannot be done with DashedVMobject().

2. You do have to specify the number of dashes.
DashedLine(), for example, can do it easily because the length of the
line is known.
If you want to setup the size of the dashes, you have to calculate the
length of the curve,
and this is not easy to do. VMobject has a get_arc_length(), but it is
just an approximation of the length.

3. DashedVMobject is a clever and nice hack.
But, IMHO, brush like properties should not be done with such a hack.
An extreme case would be just like having a ColoredVMobject or a
VMobjectWithBackground.

In a near future, I do think that it would be nice if there was some
sort of Brush class to comport
things like dashed_pattern*, stroke_color, stroke_width, fill_pattern,
background_color, stroke_style*,
line_cap*, line_join*, etc.
Those marked with * do not exist (yet). I would make a "stroke_style
patch" if Cairo had it.
I like to imagine a "caligraphic" stroke... like if painted with a brush. :-)

4. Also, since the number of dashes has to be integer in DashedVMobject
(https://numpy.org/doc/1.18/reference/generated/numpy.linspace.html),
the size of each dash is not constant during animation, even if you
keep updating
the number of dashes. The final effect is a dashed curve with dashes
that seem to be moving.

5. Maybe, not so important, it that DashedVMobject splits a curve into
dozens of tiny curves,
consuming memory and time. If the curve is moving and you want the
dash size to be kept constant,
you have to recreate a new DasehdVMobject (split the path once again)
for each update.


André Caldas.

TonyCrane:
Seems outdated. If you want, you can modify your code to fit in the new version of 3b1b/manim and open a new pull request :)

Issue #994, opened 2020-04-18
Problem with streamline


rajphys:
Hi,

I am having an issue to reproduce one of the animations with streamlines in div_curl.py. I have used the following command:

`python3.7 ./manim.py div_curl.py DefineDivergence  -pl -r 540`

The error message is 

`
  File "/home/manim/manimlib/extract_scene.py", line 155, in main
    scene = SceneClass(**scene_kwargs)
  File "/home/manim/manimlib/scene/scene.py", line 53, in __init__
    self.construct()
  File "div_curl.py", line 1459, in construct
    self.show_flow()
  File "div_curl.py", line 1490, in show_flow
    **self.stream_line_config
  File "/home/manim/manimlib/mobject/vector_field.py", line 229, in __init__
    **self.start_points_generator_config
TypeError: get_start_points() got an unexpected keyword argument 'delta_x'

`
Could you please help me with this? 

Thank you,
Raj

PR #1068, opened 2020-05-14
Add input parameters to StreamLines.get_start_points

Erik-White:
As mentioned in https://github.com/3b1b/manim/issues/994, attempting to initialise `StreamLines` with any values set in `"start_points_generator_config"` results in
```
TypeError: get_start_points() got an unexpected keyword argument 'delta_x'
```
A minimal example:
```
from manimlib.mobject.vector_field import StreamLines

stream_line_config = {
    "start_points_generator_config": {
        "delta_x": 0.1
    }
}
sl = StreamLines(lambda x: x, **stream_line_config)
```

The error is due to the initialise method of `StreamLines` attempting to call `get_start_points` with keyword arguments:
https://github.com/3b1b/manim/blob/ca0b7a6b06ff7787d1efab5ad3b28cfb88f67c3f/manimlib/mobject/vector_field.py#L228-L230
However `get_start_points` does not accept any parameters:
https://github.com/3b1b/manim/blob/ca0b7a6b06ff7787d1efab5ad3b28cfb88f67c3f/manimlib/mobject/vector_field.py#L265

 It is unclear if the configuration was handled differently in the past as `"start_points_generator_config"` has been used [in older code](https://github.com/3b1b/manim/blob/master/from_3b1b/old/div_curl.py#L1432-L1455), but no changes are evident in the git history.

This change provides `get_start_points` with parameters that default to the class config properties if no values are passed.

### DIFF
--- a/manimlib/mobject/vector_field.py
+++ b/manimlib/mobject/vector_field.py
@@ -262,18 +262,37 @@ class StreamLines(VGroup):
             )
             self.color_using_background_image(image_file)
 
-    def get_start_points(self):
-        x_min = self.x_min
-        x_max = self.x_max
-        y_min = self.y_min
-        y_max = self.y_max
-        delta_x = self.delta_x
-        delta_y = self.delta_y
-        n_repeats = self.n_repeats
-        noise_factor = self.noise_factor
-
+    def get_start_points(
+        self,
+        x_min = None,
+        x_max = None,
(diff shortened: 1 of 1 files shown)


Erik-White:
@eulertour any thoughts?

TonyCrane:
Seems outdated, so close this.

Issue #675, opened 2019-08-09
ZoomedScene： objects scale problem


BenEcon:
I have an issue when learn SlopeOfCircleExample of eoc.chapter6. The problem is that some objects of the zoomed-display area are rescaled when frame changes, but some objects remain the same scale. I know that 1b3b wants to keep some mobjects will remain the same width when frame changes, my question is how to set the width of these mobjects? The issue is attached below, any help will be appreciated.

<img width="1426" alt="Screen Shot 2019-08-09 at 9 35 22 PM" src="https://user-images.githubusercontent.com/53943014/62783356-d3359e00-baee-11e9-8daa-65555db26ade.png">

asustnt3:
I come across  this problem too.  the stroke of mobs get very thinner. 
In Essence of calculus 2, the problem seems to be solved.  but  I'm really confused about it.

PR #1073, opened 2020-05-15
Fixing a broken stroke width for ZoomedScene

ksofiyuk:
When using ZoomedScene, incorrect stroke width behavior is observed. 

I found that a special multiplier `self.get_frame_width() / FRAME_WIDTH` was added, which is supposed to keep the right stroke width. In fact, it introduces an incorrect behavior. See the example below:

before
after

Reproducible example:
```python
class Test(ZoomedScene):
    CONFIG = {
        "zoom_factor": 0.4,
        "zoomed_display_height": 5,
        "zoomed_display_width": 5
    }

    def construct(self):
        zoomed_camera = self.zoomed_camera
        zoomed_display = self.zoomed_display

        title = TextMobject("Before").scale(2.5).to_corner(LEFT + UP)

        zoomed_display.display_frame.set_color(BLUE)
        zoomed_display.to_corner(RIGHT + UP, buff=0)

        circle1 = Circle(stroke_width=6).scale(0.5)
        circle2 = Circle(stroke_width=3, color=BLUE).scale(0.25)
        rect = Rectangle(height=1.25, width=2.0, color=YELLOW, stroke_width=2).next_to(circle2, LEFT, buff=0)
        dot = Dot(fill_color=BLUE, color=WHITE, stroke_width=4).scale(2).next_to(circle1, LEFT, buff=0)
        zoomed_camera.frame.next_to(rect, LEFT)

        self.add(circle1, circle2, rect, dot, title)
        self.activate_zooming()
        self.play(self.get_zoomed_display_pop_out_animation())

        zoomed_camera.frame.generate_target().move_to(circle1)
        self.play(
            MoveToTarget(zoomed_camera.frame),
            run_time=6
        )
        self.wait(2)
```

I understand that the author of this code fragment originally wanted to keep the thickness of the lines in ZoomedScene the same as in the original scene. First of all, I think that this is an unexpected behavior as it is normal to expect a thickening of lines when zooming objects. Secondly, the originally proposed hack does not work, because the width and height of the scenes in pixels are different, so simple multiplication by `self.get_frame_width() / FRAME_WIDTH` does not restore the original stroke width. If someone wants to keep this behavior, it should be made optional.

### DIFF
--- a/manimlib/camera/camera.py
+++ b/manimlib/camera/camera.py
@@ -390,12 +390,7 @@ class Camera(object):
             self.get_stroke_rgbas(vmobject, background=background),
             vmobject
         )
-        ctx.set_line_width(
-            width * self.cairo_line_width_multiple *
-            # This ensures lines have constant width
-            # as you zoom in on them.
-            (self.get_frame_width() / FRAME_WIDTH)
-        )
+        ctx.set_line_width(width * self.cairo_line_width_multiple)
         ctx.stroke_preserve()
         return self
 


ksofiyuk:
It fixes this issue #675

ksofiyuk:
Additional clarifications: the old multiplier `self.get_frame_width() / FRAME_WIDTH` is superfluous because the zoom scale is already reflected in the get_cairo_context method by setting a new transformation matrix in [L297](https://github.com/3b1b/manim/blob/9100d0c6cb7513e1dd65f306311e8e2fabf01294/manimlib/camera/camera.py#L297).

P.S. I would also like to point out that [line 296](https://github.com/3b1b/manim/blob/9100d0c6cb7513e1dd65f306311e8e2fabf01294/manimlib/camera/camera.py#L296) with `ctx.scale(pw, ph)` is redundant, as the line following it sets a new transformation matrix, resetting the scale.

NavpreetDevpuri:
Make it optional something like `if constant_line_width == True` then ensures lines have constant width otherwise make lines looks normal.

NavpreetDevpuri:
May be add `constant_line_width ` in `CONFIG`

ksofiyuk:
This would be good, however, the current design does not allow this behavior to be easily implemented (with constant line width). 

Let's take a look at the reproducible example (rendered with a low quality config):
1. Main camera has `pixel_height = 480` and `frame_height = 7.994`.
2. Zoomed display camera has `pixel_height =  300` and `frame_height = 2`.
3. Width multiplier for the main camera is `480 / 7.994 = 60.04`.
4. Width multiplier for the zoomed camera is `300 / 2 = 150`.
5. If we multiply zoomed camera by `self.get_frame_height() / FRAME_HEIGHT = 2 /  7.994 = 0.063`, we'll get a new width multiplier `150 * 0.063 = 9.49`, which is completely wrong.

The main problem is that the zoomed display camera does not know the width and height of the main camera in pixels. So the ratio of `pixel_height/frame_height` is completely different for the main camera and the zoomed camera.

ksofiyuk:
The current solution, in which the stroke width is multiplied by `self.get_frame_height() / FRAME_HEIGHT`, is wrong and does not provide width consistency due to the reasons described above.

Width constancy can be implemented. The correct solution is to multiply stroke width in zoomed camera by `zoom_factor`, but to do so we have to throw zoom_factor into the zoomed camera class, which will require a change in the interface of the camera class.

TonyCrane:
Sorry, but this seems outdated, so closed this.

PR #1071, opened 2020-05-14
Fixed Some bugs of code_mobject.py and Text_mobject.py

NavpreetDevpuri:
Fixed https://github.com/3b1b/manim/issues/1067
NOTE : SurroundingRectangle() only contains visible text.
```Python
class te(Scene):
    def construct(self):
        text1 = Text(" ab ", font="Consolas", size=2)
        rect1 = SurroundingRectangle(text1)
        text1.chars[1].set_color(GREEN)
        self.add(text1,rect1)
```
Output
te

```python
class bug1(Scene):
    def construct(self):
        text1 = Text("  ab\ncd", font="Consolas", size=2)
        text2 = Text("ab\ngh", font="Consolas", size=2)
        rect1 = SurroundingRectangle(text1)
        rect2 = SurroundingRectangle(text2)
        self.play(Transform(remove_invisible_chars(text1), remove_invisible_chars(text2)), Transform(rect1, rect2))
        self.wait()
```
Output 
bug1

Added new parameters `background_stroke_width` and `background_stroke_color`
```python
class bug2andbug3(Scene):
    def construct(self):
        helloworldc = Code(
            "helloworldc.c",
            tab_width=4,
            background_stroke_width=1,
            background_stroke_color=WHITE,
            insert_line_no=True,
            style=Code.styles_list[4],
            background="window",
            language="cpp",
        )
        helloworldcpp = Code(
            "helloworldcpp.cpp",
            tab_width=4,
            background_stroke_width=1,
            background_stroke_color=WHITE,
            insert_line_no=True,
            style=Code.styles_list[15],
            background="window",
            language="cpp",
        )
        helloworldc.move_to(np.array([-3.6, 0, 0]))
        helloworldcpp.move_to(np.array([3.1, 0, 0]))
        self.add(helloworldc,helloworldcpp)
```
Output
bug2andbug3

Updated Paragraph() and added new methods `set_line_to_initial_position()` and `set_all_lines_to_initial_positions()`
```python
class para(Scene):
    def construct(self):
        t = Paragraph('this is a awesome', 'paragraph', 'With \nNewlines', '\tWith Tabs', '  With Spaces',
                      'With Alignments',
                      'center', "left", "right")
        t.set_line_alignment("center", 7)
        t.set_line_alignment("right", 9)
        t.shift(2 * RIGHT)
        rect = SurroundingRectangle(t)
        self.add(t, rect)
        self.wait()
        self.play(t.set_all_lines_alignments, "right")
[...]

PR #1083, opened 2020-05-18
Svgmobject update with basic processing of 'text' tag

iterater:
# Motivation

Currently,, `SVGMobject` doesn't support `<text>` tag in SVG format. The support of this tag is good for plotting diagrams, graphs, and other structures from SVG. I've faced it when trying to plot graphs rendered by [graphviz](https://www.graphviz.org/).

# Implementation

I've added text processing to SVGMobject using `Text` mobject with the following features:

- scaling text using `text_scale` for better drawing (original font size is incorrectly small;
- support of font-family translated into `font` parameter of `Text` mobject;
- additional shifting up to the half of 'o' height to adjust centered placement.

# Testing

Tested by drawing SVG file with a graph built by graphviz. The used file is available [here](https://yadi.sk/i/yvOsugyp8Cxx9g).

```python

class SvgGraphScene(Scene):
    def construct(self):
        g = SVGMobject(file_name=r'd:\g.svg', stroke_width=1.0, 
            fill_opacity=0.0, text_scale=2.0)
        g.scale(3)
        self.add(g)
        self.wait(5)
```

The result is shown below: left - before, right - after.

before_after

# Additional notes

- There are parameters in `<text>` tag which are not supported yet (further work): `text-anchor`, centering and rotation in `font-family` etc.
- It will be good to extend the logic of `SVGMobject` to support different stroke/fill styles for different objects. E.g. to fill the text instead of stroking. Currently, it can be done using `submobjects`.
- It seems that `SVGMobject` doesn't support `stroke-dasharray` in `<path>`. Could be useful. Again, currently, it seems to be available using `submobjects`.


manimlib/mobject/svg/svg_mobject.py
```diff
@@ -110,6 +113,16 @@ def get_mobjects_from(self, element):
...
 
+    def text_to_mobject(self, g_element):
+        from manimlib.mobject.svg.text_mobject import Text
+        font_size = self.attribute_to_float(g_element.getAttribute('font-size')) if g_element.hasAttribute('font-size') else 1.0
+        font_family = g_element.getAttribute('font-family') if g_element.hasAttribute('font-size') else ''
+        font_family_list = [s.strip() for s in font_family.split(',')]
+        content = g_element.firstChild.nodeValue
+        txt_mobject = Text(content, size=font_size*self.text_scale, font=font_family_list[0])
+        o_hight = Text('o', size=font_size*self.text_scale, font=font_family_list[0]).get_height() # vertical centering with 'o'
+        return txt_mobject.shift(UP*o_hight*0.5)
```
kevkevinpal, line 124: I'm getting kinda lost here so does a svg text element with the value o have incorrect vertical centering does this happen for other characters. I wonder if there is a way to do this automatically and not for the 'o' degenerate case

iterater: By default Text mobject is centered about the top of small letters (red dot below). I think the correct behavior is to use the central line of small letters (green dot).

centering

Without a half-o height shift, the result in PR looks like this (which seems to be incorrect).

shifted_center

Also, I've looked at `TextWithFixHeight` class in `text_mobject.py` for reference of text height measuring.

NavpreetDevpuri: `TextWithFixHeight` will create problems thats why i removed that in https://github.com/3b1b/manim/pull/1071

manimlib/mobject/svg/svg_mobject.py
```diff
@@ -110,6 +113,16 @@ def get_mobjects_from(self, element):
...
         return result
 
+    def text_to_mobject(self, g_element):
+        from manimlib.mobject.svg.text_mobject import Text
+        font_size = self.attribute_to_float(g_element.getAttribute('font-size')) if g_element.hasAttribute('font-size') else 1.0
+        font_family = g_element.getAttribute('font-family') if g_element.hasAttribute('font-size') else ''
+        font_family_list = [s.strip() for s in font_family.split(',')]
+        content = g_element.firstChild.nodeValue
+        txt_mobject = Text(content, size=font_size*self.text_scale, font=font_family_list[0])
+        o_hight = Text('o', size=font_size*self.text_scale, font=font_family_list[0]).get_height() # vertical centering with 'o'
```
NavpreetDevpuri, line 123: Use `'['` instead of `'o'`

NavpreetDevpuri:
I think we have to use cairo directly to do this.
M Trying.

NavpreetDevpuri:
bandicam 2020-05-19 03-30-51-296
Look at the text background which is black colored rectangle
and the position of text is the mid point of that background rectangle which is pink colored circle 
for now i don't know how to get that center for different-different fonts

iterater:
Yes, currently, the arrangement is naturally done with whole text along the center of the rectangle (see red dots below).

test_1

The desired behavior is to arrange vertically along with the small characters of text ('o' position). To count that I propose counting extra space for each character using the original cairo function `text_extents` and comparing height and vertical bearing for the whole text and separate characters. So the text can be arranged as follows. Green dots - new vertical "center".

test_2

NavpreetDevpuri:
This need more improvements but for now this is a good base.
Merge this.

NavpreetDevpuri:
I think if we store all Text elements from svg files in a list and after that just replace Text elements in svg file with cairo rendered Texts at appropriate locations according to font and style.
This seems a better approach because we don't need to rebuild text render.

NavpreetDevpuri:
But svg file is still showing in wrong way.
Reset `text_mobject.py` 
Left side is normal 
Right side is in manim 
collage

iterater:
@NavpreetDevpuri, yes, I see. The problem was in the interpretation of vertical alignment in SVG. The vertical anchor in SVG is [along the baseline, not in center](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/text-anchor). Now the text placement seems to be correct:

test_3

NavpreetDevpuri:
Good job! 
Now implement `text-anchor` shift text right or left by the half of width according to 'text-anchor' and set default value for text stroke to 0 and fill the text with color.

iterater:
@NavpreetDevpuri 

1. The horizontal alignment was updated. See below. **NB** The `text-anchor` tag was modified manually in SVG file for "A" (`text-anchor="start"`) and "B b" (`text-anchor="end"`) nodes, so they are shifted. With the original file, everything is placed fine.

test_4

2. Stroke and fill is updated for all submobjects in base `VMobject`. So, new `update_text_stroke_and_fill()` called in `__init__` was added. The default stroke is used for text color. Stroke in the text is disabled by default.

NavpreetDevpuri:
> I think if we store all Text elements from svg files in a list and after that just replace Text elements in svg file with cairo rendered Texts at appropriate locations according to font and style. Save this new generated svg file may be in new media folder `media/svg`
> This seems a better approach because we don't need to rebuild text render.
> 
> But for a base your method is just fine.
> We can improve it later.


as i said 
a better way to use `inkscape` to convert svg's text to paths automatically according to font, size ..etc it handles everything.
You can try as follow 
install inkscape then add `C:\Program Files\Inkscape\bin;` to environment variable `path`.
open terminal or cmd at 'g.svg' file's directory
then run following command 
```console
inkscape g.svg --export-text-to-path --export-plain-svg --export-filename=output.svg
```
this will create `output.svg` containing `path` instead of `text` tags

iterater:
Yes, that possibly works. But as I see there are two issues.

1. This requires additional external software to be used. As a general solution, that looks not so good. So I believe adding the functions proposed in this PR will be better.

2. In my particular case, `SVGMobject` fails in the processing of `<path>`  tag both in my branch and in master with the following error: `TypeError: add_cubic_bezier_curve_to() missing 2 required positional arguments: 'handle2' and 'anchor'`. [Here is the input SVG file](https://yadi.sk/i/5_jTz-wtY5ZhYg). And I think it should be another issue independent of this PR. (I can create an issue here for that)

All in all, for my purpose, my update seems to be working well. See an example from my current work (sorry for text in Cyrillic):

Shapes

TonyCrane:
It looks like this pull request is out of date, as `SVGobject` has been refactored in #1731 and support for text is also in the works.
Thanks for this, but I'm closing this pull request.

PR #1085, opened 2020-05-19
Remove 3b1b specific files

safinsingh:
I think we should remove 3b1b-specific files because they aren't required in this community-based repo.


NavpreetDevpuri:
This files are very helpful 
You may ask How?
You can search through those files(in pycharm press Ctrl+Shift+F) for some text or may be some function usages for just see some examples for some functions.
And you can learn so much from those files.

qustions:
@NavpreetDevpuri  nearly all files are having errors  example take
```
python3 ../../manim.py alt_calc.py  -pl
```
```
  File "Projects/editing/manim/manimlib/extract_scene.py", line 155, in main
    scene = SceneClass(**scene_kwargs)
  File "Projects/editing/manim/manimlib/scene/scene.py", line 75, in __init__
    self.construct()
  File "alt_calc.py", line 1824, in construct
    color = self.sample_dots[len(self.sample_dots) / 2].get_color()
  File "Projects/editing/manim/manimlib/mobject/mobject.py", line 913, in __getitem__
    return self_list.__getitem__(value)
TypeError: list indices must be integers or slices, not float
```

PR #1111, opened 2020-05-28
Remove unnecessary import.

teymour-aldridge:

You should also include:

The import isn't needed.
The code still runs without the import.


TonyCrane:
Already been fixed in the new version, so I'm closing this.

Issue #1103, opened 2020-05-25
Why "Mobject.get_merged_array()" function is called twice? Is it by intention?


mjmjmtl:
I just notice that the get_merged_array() function is called twice during the recursion (see line 701 and 704 below). Is line 704 redundant?
![image](https://user-images.githubusercontent.com/39451663/82782620-96366000-9e11-11ea-88b7-aac543775a73.png)
See source code here: https://github.com/3b1b/manim/blob/master/manimlib/mobject/mobject.py#L704

andre-caldas:
I had made a patch some days ago. Patch #1123.

I also think it is redundant. The result is not attributed anywhere, but the function has no "side-effects" (it does not change anything).

mjmjmtl:
Glad to hear that. Thanks @andre-caldas

Tony031218:
Fixed in #1043

PR #1123, opened 2020-06-03
Remove superfluous line.

andre-caldas:
The line removed was probably accidentally left behind.

The line calls a function with no side-effects, but its return value is not used at all.


andre-caldas:
Closes #1103.

Tony031218:
Fixed in #1043

PR #1124, opened 2020-06-03
Allow VMobject to have non-VMobject submobjects.

### LINKED ISSUE #1122: Submobjects of VMobject.
I believe, it would be a good idea if VMobject were more tolerant to having general Mobjects (non-VMobjects) as submobjects.

Rationale:
1. Mobjects, together with all its submobjects can be, for example, moved around.
2. It would be nice to being able to Mobject.add() some sort of anchors to a VMobject.
3. This way, if the vmobject is moved or rescaled, the anchors are automatically updated.
4. For example, to get a circle center, I am using VMobject.get_center(). However, during ShowCreation, VMobject.get_center() changes its position.
5. Worse than that, when you move a circle, for example, Circle:arc_center is not updated. Where it a submobject, it would be (hopefully) always updated.
6. Almost the same can be said about Line:start and Line:end.

Is it a good idea to make VMobject more tolerant to non-VMobject submobjects? I can implement it... if people think it is good.

andre-caldas:
Fixes #1122.

Rationale for this patch can be found on Issue #1122.

Calls to Mobject.move_to(), for example, end up in a call to self.get_points_defining_boundary(). However, if a VMobject contains a non-VMobject submobject, we gen an error indicating that submobject.get_anchors() is not implemented.

The method VMobject.get_points_defining_boundary() is a recursive method inherited from VMobject and reimplemented. Since get_anchors() is not implemented in Mobject, instead of calling submobject.get_anchors() for all sobmobjects, it should call get_anchors() just for self, and get_points_defining_boundary() for all submobjects.

One problem in this patch is that this implementation relies on:
1. The fact that self is the first Mobject returned by Mobject.get_family().
2. The fact that self is NOT in Mobject.get_family()[1:]

### DIFF
--- a/manimlib/mobject/types/vectorized_mobject.py
+++ b/manimlib/mobject/types/vectorized_mobject.py
@@ -137,7 +137,8 @@ class VMobject(Mobject):
     def set_fill(self, color=None, opacity=None, family=True):
         if family:
             for submobject in self.submobjects:
-                submobject.set_fill(color, opacity, family)
+                if hasattr(submobject, 'set_fill'):
+                    submobject.set_fill(color, opacity, family)
         self.update_rgbas_array("fill_rgbas", color, opacity)
         return self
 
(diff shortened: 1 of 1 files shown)


andre-caldas:
Made other VMobject methods that recurse for submobjects to ignore submobjects that are not of type VMobject.

Also, fixed the problem of possibly making an infinite recursion if self is a submobject of itself. I don't think this happens, but avoided it anyway.

3b1b:
The reason for VMobjects only having other VMobjects in their family is so that they can be rendered all together more efficiently.  If you want a group containing types that are rendered differently, the preferred approach would be to use Group(mob1, mob2, etc.).

PR #1125, opened 2020-06-04
Update graph_scene.py: Add an argument to setup the axes without displaying them on screen.

saarthdeshpande:
1. The aim / objective is to actually have a GraphScene and corresponding graphs, but not displaying the axes beforehand.

Current problem:

    You cannot add a graph without self.setup_axes()
    self.setup_axes() displays the axes by default.

For example, I want to first display some text, FadeOut the TextMobject and then FadeIn the axes, I cannot FadeIn(self.axes) unless first setting up the axes via self.setup_axes().

Hence this patch.

2. Attached is a GIF using my patch.
hideaxes_trial

### DIFF
--- a/manimlib/scene/graph_scene.py
+++ b/manimlib/scene/graph_scene.py
@@ -65,7 +65,7 @@ class GraphScene(Scene):
         self.right_T_label = VGroup()
         self.right_v_line = VGroup()
 
-    def setup_axes(self, animate=False):
+    def setup_axes(self, animate=False, hideaxes=False):
         """
         This method sets up the axes of the graph.
 
@@ -141,6 +141,8 @@ class GraphScene(Scene):
 
         if animate:
             self.play(Write(VGroup(x_axis, y_axis)))
+        elif hideaxes:
+            pass
         else:
             self.add(x_axis, y_axis)
         self.x_axis, self.y_axis = self.axes = VGroup(x_axis, y_axis)


andre-caldas:
The solution to your problem depends on what your needs are and what you are trying to accomplish. It might be a good idea to implement your patch... but I dont't know, since you didn't write anything to justify your pull request.

Maybe, instead of patching GraphScene, I really think that this VGroup(x_axis, y_axis) should simply be a separate Mobject called Axis. Maybe, you should subclass Scene and create your own MyGraphMyStyleScene.

Talking about my own experience, it took me sometime to realize that I do not need a graph scene... but having an Axis mobject would be nice. So, my question to this thread is:
- Should we have a super-customizable fully automatic GraphScene?
- Or should it be considered just a "quick" solution for a quickly done animation?
- Should we subclass GraphScene, ZoomedScene, PiCreatureScene all at the same time? Or shall we simply subclass Scene and add mobjects, cameras and animations to it?

Those are not rhetorical questions. I really wonder. :-)

I do think the patch is good and usefull... ;-)

saarthdeshpande:
The aim / objective is to actually have a GraphScene and corresponding graphs, but not displaying the axes beforehand. 

Current problem: 
- You cannot add a graph without self.setup_axes()
- self.setup_axes() displays the axes by default. 

For example, I want to first display some text, FadeOut the TextMobject and then FadeIn the axes, I cannot FadeIn(self.axes) unless first setting up the axes via self.setup_axes().

Hence this patch.

Also, a separate Mobject for the axis sounds like a great idea. A temporary fix can be drawing a graph in [draw.io](draw.io), exporting it as an SVG and using that SVG in manim as an Mobject.

As for your question, to subclass Scene and add the features would be a more feasible idea than to subclass multiple Scene(s). It would still require quite some time for the subclass patch, but will definitely be useful!

TonyCrane:
Sorry, but now `GraphScene` is no longer supported, its function has been moved to `CoordinateSystem`. So closed this

PR #1135, opened 2020-06-14
zip to list

trigaten:
https://stackoverflow.com/questions/27431390/typeerror-zip-object-is-not-subscriptable: In Python3 zip returns an iterable and thus cannot be sliced. It can be converted to a list, then sliced, however

You should also include:


TonyCrane:
All of the projects of 3b1b have been removed from this repository. So I'm closing this.

PR #1147, opened 2020-06-22
Pygments added to environment.yml

MeatBoyed:
1. After installing manim with Anaconda using the environment.yml file and running any script including the example_scenes.py script, one receives an error of "no module pygments installed" (to that effect). This dependency is in the requirements.txt file but not in the environment.yml file, therefore I added it for easier environment installation and setup.


Tony031218:
@eulertour Could you merge this pr? Many people are facing this simple problem.

PR #1159, opened 2020-07-06
Update graph_scene.py, specify the num decimal places in class GraphScene

cznull:

You should also include:

### DIFF
--- a/manimlib/scene/graph_scene.py
+++ b/manimlib/scene/graph_scene.py
@@ -34,6 +34,7 @@ class GraphScene(Scene):
         "x_leftmost_tick": None,  # Change if different from x_min
         "x_labeled_nums": None,
         "x_axis_label": "$x$",
+        "x_num_decimal_places": 0,
         "y_min": -1,
         "y_max": 10,
         "y_axis_height": 6,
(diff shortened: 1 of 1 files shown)


cznull:
i need to show a axis from 0 to 2pi, like fig.1 in http://dx.doi.org/10.1103/PhysRevLett.95.146802

code:

class Plotband(GraphScene):
    CONFIG = {
        "x_min" : math.pi*-1.0,
        "x_max" : math.pi*3.0,
        "x_axis_width": 24,
        "x_axis_label": "$k$",
        "x_tick_frequency": math.pi*0.25,
        "x_labeled_nums" :[0.0,math.pi*0.5,math.pi*1.0,math.pi*1.5,math.pi*2.0],
        "x_num_decimal_places": 2,
        "y_min" : -2.0,
        "y_max" : 2.0,
        "y_axis_height": 8,
        "y_axis_label": "$E$",
        "y_tick_frequency": 0.5,
        "y_labeled_nums" :[-2.0,-1.0,0.0,1.0,2.0],
        "y_num_decimal_places": 0,
        "graph_origin" : ORIGIN ,
        "function_color" : RED ,
        "axes_color" : GREEN,
        "graph_origin": [-6,0,0]
    }   
    def construct(self):
        self.setup_axes(animate=True)

3b1b:
The current plan is to deprecate GraphScene and have all that functionality live in CoordinateSystem, in which case configuration for the axes labels should happen in Axes.add_coordinate_labels.  Right now it looks like that method does not let you pass any configuration for the numbers, but I can change that.

PR #1170, opened 2020-07-15
Update complex_transformation_scene.py #498

yinjann:
Fixing problems with the animation of the zeta transformation

You should also include:

### DIFF
--- a/manimlib/once_useful_constructs/complex_transformation_scene.py
+++ b/manimlib/once_useful_constructs/complex_transformation_scene.py
@@ -125,11 +125,9 @@ class ComplexTransformationScene(Scene):
         transformer, transform_kwargs = self.get_transformer(**kwargs)
         transformer.generate_target()
         # Rescale, apply function, scale back
-        transformer.target.shift(-self.background.get_center_point())
+
         transformer.target.scale(1. / self.background.unit_size)
-        transformer.target.apply_complex_function(func)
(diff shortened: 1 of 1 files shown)


giladsofer:
Sorry if this is basic, but how do you install these updates? (this and also update coordinate system)

yinjann:
You don't have to apologize. click on my pull request "Update complex_transformation_scene.py" there you will see the changes you have to make. Basically it's going to the file manim / manimlib / once_useful_constructs / complex_transformation_scene.py
there you have to delete the lines 128,129,131,132
This was the cause of the animation acting strangely. I suppose it was because they later updated other parts of the code and those lines were left over

giladsofer:
I deleted the lines you suggested and although the animation is a bit better, there is still a lot of noise due to the singularities of the function (I tried it with f(z)=1/z). Do you have an idea on how to solve this?

yinjann:
yes, it seems that add_extra_plane_lines_for_zeta() doesn't work well.
Instead use: 

        prueba = self.get_dense_grid()
        self.plane.submobjects = [prueba]
        self.apply_zeta_function()

Check my changes #1169

giladsofer:
But is this relevant for the function 1/z if I'm not using the zeta function? (I only use complex transformation scene)

lemsi19 (commented):
1b82ea85074548f2ae1e8f9320bdba0a52760c49!

PR #1184, opened 2020-07-21
Physic

pdcxs:
Add physics simulation support by add package 'pymunk'.

### DIFF
--- a/example_scenes.py
+++ b/example_scenes.py
@@ -133,4 +133,73 @@ class UpdatersExample(Scene):
         )
         self.wait()
 
+class PhysicExample(PhysicScene):
+    def construct(self):
+        self.set_gravity(9.8 * DOWN)
+
+        grd_body = self.space.static_body
+        grd_seg = pymunk.Segment(grd_body, (-10, -2), (10, -2), 0)
+        grd_seg.friction = 0.7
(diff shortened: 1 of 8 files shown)


3b1b:
Unfortunately, this is now based on an older version of manim.  One difference is that there shouldn't be a need to do have the plot_depth attribute which the camera interacts with, but for mobjects that should render in the correct order based on where they are in 3d space you should be able to call mobject.apply_depth_test.

Rather than writing an Animation for Simulate here, it might be better to do this with Mobject updaters.  Animations are designed to take place over finite periods of time, whereas an updater is meant for changes happening perenially.  I might also suggest trying to have as much functionality isolated to being part of a special Mobject, rather than a special Scene.  For example, you might define a Simulation mobject with an updater that determines how all its submobjects evolve according to physical laws.  This way, it's easier to add the Simulation as just a part of another scene where you might want to have other animations going on simultaneously, say indicating parts of what's going on in the simulation, or have two separate ones run side-by-side.

If it's not too much trouble to adapt it so that there are no conflicts with the current version, feel free to re-open the PR.  Also, you may want to consider it as a contribution to the [community edition](https://github.com/ManimCommunity/manim).

PR #1200, opened 2020-08-09
Updated line_intersection code

raviarora1209:
Changed the line__intersection

Earlier, it used the array system of marking the points of lines which returns an error:-
image

Now, it works perfectly:-
image

### DIFF
--- a/manimlib/utils/space_ops.py
+++ b/manimlib/utils/space_ops.py
@@ -215,8 +215,8 @@ def line_intersection(line1, line2):
     each defined with a pair of vectors determining
     the end points
     """
-    x_diff = (line1[0][0] - line1[1][0], line2[0][0] - line2[1][0])
-    y_diff = (line1[0][1] - line1[1][1], line2[0][1] - line2[1][1])
+    x_diff = (line1.get_start()[0] - line1.get_end()[0], line2.get_start()[0] - line2.get_end()[0])
+    y_diff = (line1.get_start()[1] - line1.get_end()[1], line2.get_start()[1] - line2.get_end()[1])
 
     def det(a, b):
         return a[0] * b[1] - a[1] * b[0]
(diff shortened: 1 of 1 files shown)


Tony031218:
this function is used as `line_intersection(line1.get_start_and_end(), line2.get_start_and_end())` where `line1, line2` are two `Line` object. So there may be no wrong with the code.

raviarora1209:
@Tony031218 Well, I wasn't aware of that. Thanks for the insight, but don't you think it would be easier to use if the people just had to pass the `Line` object rather than personally having to access it...Not saying that it would bring drastic changes to the workflow, but it could help the beginners in being able to easily access the functions just by looking at their code or seeing it in auto-complete recommendations. Also, since I am new to the Manim community, do you also accept new vital functions or scene types as PRs? Like a function for the intersection of two circles and or a template scene for light reflection and all..

Tony031218:
Yes, it is really inconvenient to use this way at present, I was also a little troubled when I first used it.
But if you change it according to your method, then the usage of this function will be completely changed, causing some old sample scenes to be completely unusable.

So I suggest you judge whether the incoming `line1, line2` are two `Line` instances at the beginning of the function, and if so, use your method.  This can support two incoming methods at the same time. 
And the original usage has an advantage that you don’t need to create a `Line` when you want to get an intersection, but directly pass in an iterable type containing two coordinates, such as `line_intersection((LEFT, RIGHT), (UP, DOWN))`.  This may be more convenient at some point.

For community contributions, I do not have the permission to merge pull requests (but I am striving for it).  The functions you mentioned may not be easy to implement, so there are currently no such functions.  But you are always welcome to contribute codes for them.  And, [ManimCommunity](https://github.com/ManimCommunity/manim) maybe more welcome you to add these features.

raviarora1209:
Thanks a bunch @Tony031218...I really appreciate your help and assistance and thanks for all the insight you provided..although maybe I still could rename the function and probably wait for the PR.

PR #1204, opened 2020-08-16
help(Scene.play)

parmentelat:

trying to get help on `Scene.play` leads to a dead end, because the `handle_play_like_call` decorator fails to take care of its outcome's docstring (and other special attributes like `__name__` I would bet)

the good practice for writing such decorator is to take advantage of `functools.wraps`, this PR exhibits one example; let me know what you think and I could review the code more deeply to account for the other possible places where that change could come in handy

this way, decorated methods - typically `Scene.play` - have a decent docstring
so first-time users can use `help()` on them

well, I'm a first time user, and I've been using this brew of manim for my first steps, and have not met any unexpected result so far, as compared to what the tutorial shows

### DIFF
--- a/manimlib/scene/scene.py
+++ b/manimlib/scene/scene.py
@@ -2,6 +2,7 @@ import inspect
 import random
 import warnings
 import platform
+from functools import wraps
 
 from tqdm import tqdm as ProgressDisplay
 import numpy as np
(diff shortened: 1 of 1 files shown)


3b1b:
Thanks for letting me know, I did not know the proper convention here.  Manim has changed a lot since then, would it be too much to ask that you reopen the PR for the latest version?

PR #1214, opened 2020-08-26
Fixed import errors with diffyq files in active_projects directory

tj-murphy:
From a fresh Manim install, attempting to execute 
`python3 ./manim.py ./from_3b1b/active/diffyq/all_part1_scenes.py ` 
results in an error because the first line of `all_part1_scenes.py` says `from active_projects.diffyq.part1.pendulum import *`

This doesn't work, since the "active_projects" directory doesn't exist in the Manim repo. Rather, this directory is named "active".

Simply changing the script to use "active" instead fails - I'm not sure why this is, but going up another level to the from_3b1b directory fixes everything. I am now able to execute the diffyq files.


Tony031218 (approved):

TonyCrane:
All the projects of 3b1b have been removed from this repository. So I'm closing this.

PR #1215, opened 2020-08-27
added a 3d cone and a 3d arrow

dor2727:
Checked.

Adding more 3d objects - a Cone, and a 3d arrow.

minimal working example:

[cone_example.zip](https://github.com/3b1b/manim/files/5135704/cone_example.zip)


3b1b:
Unfortunately, this is now based on an outdated version of manim.  In the new version, there is a dedicated surface type which is made for OpenGL, so these additions might work nicely in that context.  Take a look at manimlib/mobject/three_dimensions.py

PR #1240, opened 2020-09-29
Replace os.system(...) with subprocess.call(...)

YuanZ1949:
在某些情况下 `os.system` 无法完成特定的逻辑
In some cases, `os.system` cannot complete the specified logic 

比如:  `os.system(" ".join(commands))` 就不能正确的执行
e.g `os.system(" ".join(commands))` will not execute correctly

我们将无法创建 `.xdv` , `.svg` 文件
We will not create `.xdv` , `.svg` file

使用 `subprocess.call(...)` 替换 `os.system(...)` 
Replace os.system(...) with subprocess.call(...)

[replacing-os-system@Python3/library](https://docs.python.org/3/library/subprocess.html#replacing-os-system)


Tony031218:
注意 Note that:
```text
xelatex: major issue: So far, no MiKTeX administrator has checked for updates.
```
所以我认为使用管理员权限，并且检查MiKTeX的更新是你需要解决的主要问题。
So I think using administrator privileges and checking for MiKTeX updates are the main problems you need to solve.

在正常情况下，对于这里的用法，`os.system()`和`subprocess.call()`不应该有区别，并且我也没能复现你的报错。所以我觉得你应该修复你本机的问题，而不是更改这里的源码。
Under normal circumstances, for the codes here, there should be no difference between ʻos.system()` and `subprocess.call()`, and I have not been able to reproduce your error. So I think you should fix the problem of your computer instead of changing the source code here.

YuanZ1949:
多谢回复
Thanks for reply

通过一些控制变量的方法, 大概有了基本的结论:
Through some methods of controlling variables, I have probably reached a basic conclusion:

- 如果是 `MiKTeX`的问题的话, `dvisvgm` 应该可以正常工作才对. 所以按照提示更新后依然如此:
   If it is a problem with `MiKTeX`, `dvisvgm` should work normally. So it is still the same after updating as prompted:

```cmd
❯ python -m manim example_scenes.py WriteStuff -pl
Media will be written to ./media\. You can change this behavior with the --media_dir flag.
Writing "\centering This is a some text" to ./media\Tex\4475e3a270ecfb06.tex

  File "D:\manim\manimlib\extract_scene.py", line 155, in main
    scene = SceneClass(**scene_kwargs)
  File "D:\manim\manimlib\scene\scene.py", line 75, in __init__
    self.construct()
  File "example_scenes.py", line 102, in construct
    tex_to_color_map={"text": YELLOW}
  File "D:\manim\manimlib\mobject\svg\tex_mobject.py", line 148, in __init__
    self, self.arg_separator.join(tex_strings), **kwargs
  File "D:\manim\manimlib\mobject\svg\tex_mobject.py", line 44, in __init__
    self.template_tex_file_body
  File "D:\manim\manimlib\utils\tex_file_writing.py", line 21, in tex_to_svg_file
    dvi_file = tex_to_dvi(tex_file)
  File "D:\manim\manimlib\utils\tex_file_writing.py", line 72, in tex_to_dvi
    "See log output above or the log file: %s" % log_file)
Exception: Latex error converting to dvi. See log output above or the log file: media/Tex/4475e3a270ecfb06.log
```

所以如上面的报错, 我碰到的问题应该不是 `Note` 给出的问题, 不过感谢提醒, 更新后没有多余的提示也是好的.
So like the above error, the problem I encountered should not be the problem given by `Note`, but thanks for the reminder, it is good that there is no extra prompt after the update.

- 在我使用的最初的manim版本 `commit`: [a42d9c6ecb4f94542668651b5fe912a5cf9ec032](https://github.com/YuanZ1949/manim/commit/a42d9c6ecb4f94542668651b5fe912a5cf9ec032), 应该和我使用的终端有未知的调用不正确的情况. 我用的是 `Window Terminal` + `cmd&pwsh` + `Anaconda` , 在实际的其他工作中, 也出现过直接运行`cmd`然后调用命令可以工作, 但是使用`Window Terminal`等工具无法得到同样结果的情况. **我觉得是** `Window Terminal` + `cmd&pwsh` **这套工具的** `bug` (甚至有时候重开终端就又可以得到正确的结果)

  In the initial version of manim I used  `commit`: [a42d9c6ecb4f94542668651b5fe912a5cf9ec032](https://github.com/YuanZ1949/manim/commit/a42d9c6ecb4f94542668651b5fe912a5cf9ec032), there should be an unknown  incorrect call with the  terminal I am using. It is `Window Terminal` + `cmd&pwsh` + `Anaconda`. In other actual work, there have been cases where running `cmd` directly and then calling the command can work, but using tools such as `Window Terminal` can not get the same result. **I think it is the** `bug` **of the tool** `Window Terminal` + `cmd&pwsh` (Sometimes you can get the correct result by restarting the terminal)

- 通过 `merge` 后面由你 `merge` 的 `pull requests` , 在不修改 `os.system()` 的情况下, 我用的这套工具也可以工作了(反而改成 `subprocess.call()` 出现过一次奇怪的失败, 重开终端后正常). 在一系列的最近 `merge` 的 `pull requests` 里, 可能有刚好解决这个未知 `bug` 的更好的解决方案.
  Through the `pull requests` of `merge` followed by `merge`, without modifying the ʻos.system()`, the tool I used can also work (instead it is changed to `subprocess.call()` There was a strange failure, and it was normal after reopening the terminal). In a series of recent `pull requests` of `merge`, there may be a better solution to this unknown `bug`.

总之, 不修改应该还是更好的吧. 
In short, it should be better not to modify it.

主要的导致`Window Terminal` 不能正常完成工作的未知 `bug` 已经被某个 `pull request` 解决了.
The main unknown `bug` that caused `Window Terminal` to not complete the work normally has been resolved by a `pull request`.

非常感谢你的工作
Thank you very much for your work

PR #1247, opened 2020-10-01
Added a method for fixed in frame labels

nutanstrek:
Hi, 
In my latest project i used axes labels which always face towards the camera and they also move
with the tips of the axes so I added a method to the SpecialThreeDScene class simply because my 
method "get_fixed_in_frame_labels" uses ThreeDScene class.

Here is a video which proves my point : 

ezgif com-gif-maker

### DIFF
--- a/manimlib/scene/three_d_scene.py
+++ b/manimlib/scene/three_d_scene.py
@@ -350,72 +350,69 @@ class SpecialThreeDScene(ThreeDScene):
         self.set_camera_orientation(
             **self.default_angled_camera_position
         )
-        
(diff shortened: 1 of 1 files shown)


3b1b:
Unfortunately, this is based on a now-outdated version.

In the newer version, being fixed in a frame is handled by the mobject, not the camera, and in particular in get_gl_Position.glsl.  This would actually be a little tricky to implement in a way analogous to how Mobject.fix_in_frame now does it, but it might be possible to add an updating function that ensures the unit normal of a mobject is always pointed toward a given point in space (e.g. that of the camera).

PR #1259, opened 2020-10-19
Update learning_by_example.rst

GauravSarkar:
corrected grammatical errors. A command was not clear so modified it.

You should also include:


Tony031218:
Perhaps no need to update docs now. So closing this.

PR #1264, opened 2020-11-04
Specify pycairo version to 1.19.1

niedong:
pip will by default install the newest version of library. The newest pycairo(which manimlib needs) is now version [1.20.0](https://pypi.org/project/pycairo/#history) released on October 6th, 2020. But unfortunately pycairo 1.20.0 uses [PEP 517](https://www.python.org/dev/peps/pep-0517/) to build, which still has problems with virtualenv that travis-ci or other users rely on.

See the current travis-ci output [message](https://travis-ci.org/github/3b1b/manim/jobs/737736332#L377).

The second newest pycairo version on pip now is 1.19.1, and will not cause the ci to fail. In the future, new version of pycairo should always be tested to see if it's working, otherwise it's safer to specify a particular version of pycairo that works.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ progressbar
 scipy
 tqdm
 opencv-python
-pycairo
+pycairo==1.19.1
 pydub
 pygments
 pyreadline; sys_platform == 'win32'


niedong:
See this pull request testing result [here](https://travis-ci.org/github/3b1b/manim/pull_requests).

naveen521kk:
You guys are thinking about CI and simply not thinking about the users. Pycairo version 1.20.0 has wheels for windows build, which is much required for Windows Users which isnt there in other version.

niedong:
> You guys are thinking about CI and simply not thinking about the users. Pycairo version 1.20.0 has wheels for windows build, which is much required for Windows Users which isnt there in other version.
> 
> And also, seeing the error message it clearly, says to update the system library cairo, which it found to be less then 1.15.0, That can be done with just another line of code and not doing this.
> CC
> @Tony031218 @niedong

Thanks.
I found where the problem is. I'll create a new branch and make a pull a request.  This pr is gonna be closed. Let me know if there's still issue in the new pull request.

Tony031218:
> You guys are thinking about CI and simply not thinking about the users. Pycairo version 1.20.0 has wheels for windows build, which is much required for Windows Users which isnt there in other version.
> 
> And also, seeing the error message it clearly, says to update the system library cairo, which it found to be less then 1.15.0, That can be done with just another line of code and not doing this.
> CC
> @Tony031218 @niedong

Sorry for that. That's my fault.

PR #1274, opened 2020-11-18
improve image background

lovemyliwu:

You should also include:


3b1b:
Unfortunately, it looks like this is a little outdated based on the current version of manim, so I'm closing it.

PR #1301, opened 2021-01-02
Some cleanups and bug fixes for the shaders branch

Tony031218:
Here are some cleanups and bug fixes for the shaders branch

- Deleted some unnecessary files and folders:
   - The documents in `docs/` are for the master branch and are not currently being updated, so delete them in the shaders branch
   - Travis related is mainly used to build documentation and publish python package, which are useless in shaders branch, delete
   - The docker image is for the master branch, not for the shaders branch, so delete
- Deleted the version of the package in `requirements.txt` for easy installation, and updated the `environment.yml` for anaconda
- Fixed the `TypeError` caused by float number when determining the window position
- Fixed the code that may cause `(xe)Latex error converting to dvi(xdv)`, same as #689 #1187
- Fixed the bug in `extract_scene.py`, otherwise will cause the error of `TypeError: run() missing 1 required positional argument:'self'`after selecting the scene number
- Updated `README.md`


3b1b:
Looks good to me, thanks!

NeoPlato:
What about IPython as a dependency mentioned in `manimlib\scene\scene.py`

Could that be added as well?

Tony031218:
> What about IPython as a dependency mentioned in `manimlib\scene\scene.py`
> 
> Could that be added as well?

Added it in #1315, thanks for that.

PR #1315, opened 2021-01-15
Fix some bugs of the shaders branch

Tony031218:
Here are some bug fixes for the shaders branch.

- The first commit https://github.com/3b1b/manim/commit/79b81363309d4b9eb1a125d6dac75ee5cd674bc1 : I'm trying to make the `Text` class work perfectly in shaders. But I found that if I remove some of the changes in #1030(or #1031), the `Text` will directly work fine. So I revert them.
- The second commit https://github.com/3b1b/manim/commit/78848b971cd3f7db0831db58ad4e9c40d1cf61e4 : I checked the docs of `logging` in the official python documentation and also checked the source code. I found that `logging.log` cannot receive the `file` argument, so I deleted this line, otherwise the following error will appear:
```python
  File ".../manimlib/extract_scene.py", line 82, in get_scenes_to_render
    file=sys.stderr
  File ".../Anaconda3/envs/manim/lib/logging/__init__.py", line 2014, in log
    root.log(level, msg, *args, **kwargs)
  File ".../Anaconda3/envs/manim/lib/logging/__init__.py", line 1449, in log
    self._log(level, msg, args, **kwargs)
TypeError: _log() got an unexpected keyword argument 'file'
```
- The fourth commit https://github.com/3b1b/manim/commit/f696ba4100d43e67a3ebd05c05b514086cf435b6 : I write a brief example for `Text` to introduce and test it.
- The fifth commit https://github.com/3b1b/manim/pull/1315/commits/fa3cad18e0e4d55e727b09ff210a79befa612b19 : Added the missing `ipython` package mentioned in https://github.com/3b1b/manim/pull/1301#issuecomment-753643071, and updated the `environment.yml` for Anaconda.
- The sixth commit https://github.com/3b1b/manim/pull/1315/commits/c6fcaa93bb9c4482815fd03b19de2a52951cf143 : Deleted the file `stream_starter.py`. Because it seems that stream related functions have been completely deleted, but this file remains.
- The eighth commit https://github.com/3b1b/manim/pull/1315/commits/7c04edcced38eea1046b934aa853e8500c4b3af9 : Added `font_size` to `Text`, and synchronized to the same size effect as `TextMobject` :
VG))B17{8$14CRJGRMDD7`R

---
@3b1b please have a look at these changes.
By the way, there is also a bug that I can't fix it, I talked in detail in #1314

PR #1316, opened 2021-01-16
Improve Text's handling of non-display characters in shaders and Add a feature

TonyCrane:
Continued https://github.com/3b1b/manim/pull/1315#issuecomment-761285165
Although `Text` can be colored properly in some cases, the position of non-display characters is not very good, and tabs are not processed. Because the #1035 of the master is not synchronized to the shaders. So I added them (https://github.com/3b1b/manim/commit/5f6b653803e3b138a2c9337c4a22dead375299c8) and made some changes (https://github.com/3b1b/manim/commit/3d118a5bb6d0dd55111dc8bc496db07d010f26d9) to fit the shaders branch.
And for this part (↓), if modify `SVGMobject` as in #1035, it may cause some problems for SVG, so I only modified the generated svg file in `Text` and deleted the empty path.
https://github.com/3b1b/manim/blob/5f6b653803e3b138a2c9337c4a22dead375299c8/manimlib/mobject/svg/text_mobject.py#L92-L97
Now, even **tabs**, **newlines**, and different amounts of **whitespaces** will not affect the coloring:
```python
text = Text(
    """Here is a tab ->\t, and a newline ->\nand even the number of whitespaces    doesn't influence coloring""",
    t2c={"tab": BLUE, "newline": BLUE, "number": BLUE, "coloring": ORANGE}
).set_width(FRAME_WIDTH - 1)
self.add(text)
index = index_labels(text, 0.05)
self.add(index)
```
E88({N`~C0M9{ 0VZD8J3WH

And one of the differences between `Text` and `TextMobject` is that the index of `Text` completely obeys the incoming string, that is, tabs are reserved (each occupies a position), and whitespaces (each occupies a position) are also reserved. I think this may be convenient for using `Text`.

---

- For this commit https://github.com/3b1b/manim/commit/f6b5edede29a972c2a1b08d035b37119a0a9d2c0, if don't modify it in this way, the computer will automatically select the Chinese input method when I (in China) focus on the window, resulting in the following annoying errors repeatedly. After the modification, only one sentence is prompted, which may be more friendly.
```text
  File "_ctypes/callbacks.c", line 232, in 'calling callback function'
  File "...\Anaconda3\envs\manim\lib\site-packages\pyglet\window\win32\__init__.py", line 668, in f
    result = event_handler(msg, wParam, lParam)
  File "...\Anaconda3\envs\manim\lib\site-packages\pyglet\window\win32\__init__.py", line 739, in _event_key
    self.dispatch_event(ev, symbol, modifiers)
  File "...\Anaconda3\envs\manim\lib\site-packages\pyglet\window\__init__.py", line 1327, in dispatch_event
    if EventDispatcher.dispatch_event(self, *args) != False:
  File "...\Anaconda3\envs\manim\lib\site-packages\pyglet\event.py", line 408, in dispatch_event
    if handler(*args):
  File "...\manimlib\window.py", line 66, in on_key_press
    self.scene.on_key_press(symbol, modifiers)
  File "...\manimlib\scene\scene.py", line 542, in on_key_press
    if chr(symbol) == "r":
OverflowError: Python int too large to convert to C long
```

- For this commit https://github.com/3b1b/manim/commit/b7bd40a6f6af9ec56237529ba954935c62b51d12, I added an option to change the position of the window in `custom_default.yml`. You can use the relative position on the screen (the first character means upper(U) / middle(O) / lower(D), the second character means left(L) / middle(O) / right(R)), and you can also use the pixels on the screen in the upper left corner of the window (separated by a comma) for positioning.

---
Btw, you have gave me the permission to directly push into the shaders branch, but in order not to interfere with your work, I did not do this, but opened a pull request and waited for you to read it and merge. 
So now if I find a small problem in codes (like https://github.com/3b1b/manim/pull/1316/commits/cfa8577454c5272ccad3d7103e710e9d17ae1643), I can _directly commit and then push into the shaders branch_, or _open a pull request but merge directly_, or _open the pull request and wait for you to merge_. Which one do you prefer? @3b1b

### DIFF
--- a/custom_defaults.yml
+++ b/custom_defaults.yml
@@ -29,6 +29,10 @@ universal_import_line: "from manimlib.imports import *"
 style:
   font: "Consolas"
   background_color: "#333333"
+# Set the position of preview window, you can use directions, e.g. UL/DR/OL/OO/...
+# also, you can also specify the position(pixel) of the upper left corner of 
+# the window on the monitor, e.g. "960,540"
+window_position: UR
 camera_qualities:
   low:
     resolution: "854x480"
--- a/environment.yml
+++ b/environment.yml
@@ -24,4 +24,5 @@ dependencies:
     - pyyaml
     - validators
     - ipython
+    - PyOpenGL
 
--- a/example_scenes.py
+++ b/example_scenes.py
@@ -125,6 +125,7 @@ class TextExample(Scene):
 
         fonts = Text(
             "And you can also set the font according to different words",
+            font="Arial",
             t2f={"font": "Consolas", "words": "Consolas"},
             t2c={"font": BLUE, "words": GREEN}
         )
--- a/manimlib/config.py
+++ b/manimlib/config.py
@@ -203,7 +203,23 @@ def get_configuration(args):
     else:
         window_width = monitor.width / 2
     window_height = window_width * 9 / 16
-    window_position = (int(monitor.width - window_width), 0)
+    custom_position = custom_defaults["window_position"]
+    if "," in custom_position:
+        posx, posy = map(int, custom_position.split(","))
+    else:
+        if custom_position[1] == "L":
+            posx = 0
+        elif custom_position[1] == "O":
+            posx = int((monitor.width - window_width) / 2)
+        elif custom_position[1] == "R":
+            posx = int(monitor.width - window_width)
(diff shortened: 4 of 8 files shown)


3b1b:
Looks good!

PR #1323, opened 2021-01-28
Interactive Mobjects in Manim

sahilmakhijani:
# Motivation
Having sliders, textbox, checkbox, etc will speed up the development process of videos and it can also help in some videos that include explanations of how things change when some parameters are tweaked.

# Example
InteractiveMobjectsExample
( Source code of the above interactive is added in **example_scenes.py** )

# Uses

* It can be used during development purposes for tweaking parameters and seeing the changes live.
* It can also be used for explanation videos that consist of tweaking parameters.
* For videos, instead of mouse and keyboard, you can change Interactive Mobject Values using the **set_value** method.
* Like ValueTracker, you can add updaters to other mobjects which changes mobject state using interactive mobjects **get_value** method.

# Files Changed

1. manimlib/mobject/mobject.py
> Added event handlers methods and "listen_to_events" config in Mobject Class.
2. manimlib/scene/scene.py
> Added calls to event handlers of Mobject whose "listen_to_events" property is True.
3. manimlib/mobject/interactive.py
> Added some interactive Mobjects in this file.
> This file includes
> * MotionMobject (This Mobject can be dragged to any location on-screen)
> * Button (You can pass any Mobject associate with it an on_click event listener)
> * ControlMobject (It extends ValueTracker Mobject and is base for other controls below)
> * EnableDisableButton and Checkbox (It stores bool value and toggles it on press)
> * LinearNumberSlider (It stores number value and can be changed using slider)
> * ColorSliders (It is a Group of 4 Sliders for RGBA and one preview box)
> * Textbox (It stores string value and can be changed by clicking on textbox once to activate it and typing)
> * Control Panel (It can take Mobjects and put it in a drawer on the top left corner which can be dragged down as in the example)
4. manimlib/utils/space_ops.py
> Added one space util method used for LinearNumberSlider Mobject.
5. manimlib/imports.py
> Added manimlib/mobject/interactive.py import in global imports.
6. example_scenes.py
> Added the above example source code in this file.

PR #1324, opened 2021-01-29
Interactive Mobjects Performance Improvements

sahilmakhijani:
# Motivation
As mentioned in https://github.com/3b1b/manim/pull/1323#issuecomment-769350572, interactive mobjects were a bit laggy.

So, to improve interactively mobjects performance, the following changes are made.
1. Maintain a separate list of mobject listeners instead of filtering mobjects on every frame.
>   This significantly improved the performance on all events except the on_mouse_drag event. So to improve it the 2nd change was made.
2.  Changed the way drag events are handled.
> This improved the performance of on_mouse_drag event.

# Files Changed
1. manimlib/scene/scene.py
> To store a separate list of mob_listners rather than checking it on every frame and changed the on_mouse_drag event handling. 
2. manimlib/mobject/interactive.py
> To adapt to the changes made in the way event handling works

# Change in how on_mouse_drag works
If mouse is being draged, following events are triggered
1. **on_mouse_press** once
2. Then, till the mouse is being dragged, **on_mouse_drag** is called
3. Finally, **on_mouse_release** once

Previously, on_mouse_drag event of mobject was called on every frame if the point is touching the mobject.

So, if in between dragging, the mouse point leaves the mobject radius, on_mouse_drag event of mobject is stopped.

Now, If on_mouse_press is called, all the mobjects near mouse point is stored in array. And when mouse is being dragged,
on_mouse_drag is called on all the mobjects stored earlier regardless of whether mouse point is touching the mobject or not.
And, once on_mouse_release is called, that array is cleared

This has significantly improved the interactivity involving dragging.

# Results
DraggableMobjectsExample
Now, all the interactive mobjects are responding very quickly to events.


TonyCrane:
The same as #1326? So close this

sahilmakhijani:
> The same as #1326? So close this

Forgot to close this!

PR #1325, opened 2021-01-31
OpenGL rendering

3b1b:
For a while, there has been divergent development on this branch, "shaders", aimed at shifting manim to build on OpenGL, via the library moderngl.  This makes things run real time, and introduces an interactive window to help with development and presentations.

This also includes meaningful changes in the underlying structure of Mobjects, Scenes and Animations, along with many shifts in the syntax.

The old version is still available on the branch cairo-backend.


TonyCrane (approved):

TonyCrane:
Maybe we need to [draft a release](https://github.com/3b1b/manim/releases/new) for the current `master` branch and release it to pypi as the last stable version using CPU.

3b1b:
That seems like a good idea.  I just made a cairo-backend branch off of master, shall we point the release to that?

TonyCrane:
> That seems like a good idea. I just made a cairo-backend branch off of master, shall we point the release to that?

Yes, but I'm not sure whether pypi support to do this or nor.
image

PR #1332, opened 2021-02-02
Update README.md

asutoshgha:
added installation guide for macOS X

You should also include:

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,15 @@ pip install -r requirements.txt
 # Try it out
 python manim.py example_scenes.py OpeningManimExample
 ```
-
+### Mac OSX
+1. Install [FFmpeg],[LaTeX],[cairo](if you only want to run on CPU) in commandline using homebrew.
+  ```sh
+   brew install ffmpeg mactex cairo
+   ```
+2. Install manim using this command if you dont want to tinker with code.(virtualenv wrapper is recommended)
+  ```sh
+   pip3 install manimlib
+   ```
 ### Directly (Windows)
 1. [Install FFmpeg](https://www.wikihow.com/Install-FFmpeg-on-Windows).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,7 +28,15 @@ pip install -r requirements.txt
 # Try it out
 python manim.py example_scenes.py OpeningManimExample
 ```
-
+### Mac OSX
+1. Install [FFmpeg],[LaTeX],[cairo](if you only want to run on CPU) in commandline using homebrew.
```
TonyCrane, line 32: Now cairo is only used in manim to make svg files for `Text`, and will not be used for rendering. OpenGL will be used anyway.
And adding square brackets to these dependencies is not a markdown syntax, it will show these [], which is not what we want.

README.md
```diff
@@ -28,7 +28,15 @@ pip install -r requirements.txt
...
+  ```sh
+   brew install ffmpeg mactex cairo
+   ```
+2. Install manim using this command if you dont want to tinker with code.(virtualenv wrapper is recommended)
+  ```sh
+   pip3 install manimlib
```
TonyCrane, line 38: Please do not use pip to install `manimlib` now. The installed version is still the old version in the `cairo-backend` branch. The new version in the `master` branch has not been released to pypi. Currently, manim can only be used through cloning this repository.
Moreover, it is very inappropriate to introduce the installation method of the old version in the `cairo-backend` branch in this new version README.

TonyCrane (changes requested):

README.md
asutoshgha: oh ,I didn't knew about that .I found in some tutorial that manimlib can be installed from pip but I didn't knew it was an older version.Thanks for pointing out.

PR #1326, opened 2021-01-31
Interactive Mobjects Performance Improvements

sahilmakhijani:
# Motivation
As mentioned in https://github.com/3b1b/manim/pull/1323#issuecomment-769350572, interactive mobjects were a bit laggy.

So, to improve interactively mobjects performance, the following changes are made.
1. Maintain a separate list of mobject listeners instead of filtering mobjects on every frame.
>   This significantly improved the performance on all events except the on_mouse_drag event. So to improve it the 2nd change was made.
2.  Changed the way drag events are handled.
> This improved the performance of on_mouse_drag event.

# Files Changed
1. manimlib/scene/scene.py
> To store a separate list of mob_listners rather than checking it on every frame and changed the on_mouse_drag event handling. 
2. manimlib/mobject/interactive.py
> To adapt to the changes made in the way event handling works
3. example_scenes.py
> Commented out "self.embed()"

# Change in how on_mouse_drag works
If mouse is being draged, following events are triggered
1. **on_mouse_press** once
2. Then, till the mouse is being dragged, **on_mouse_drag** is called
3. Finally, **on_mouse_release** once

Previously, on_mouse_drag event of mobject was called on every frame if the point is touching the mobject.

So, if in between dragging, the mouse point leaves the mobject radius, on_mouse_drag event of mobject is stopped.

Now, If on_mouse_press is called, all the mobjects near mouse point is stored in array. And when mouse is being dragged,
on_mouse_drag is called on all the mobjects stored earlier regardless of whether mouse point is touching the mobject or not.
And, once on_mouse_release is called, that array is cleared

This has significantly improved the interactivity involving dragging.

# Results
DraggableMobjectsExample
Now, all the interactive mobjects are responding very quickly to events.

PR #1335, opened 2021-02-02
Refactored Event Handling Mechanism

sahilmakhijani:
# Motivation
As metioned by @3b1b https://github.com/3b1b/manim/pull/1326#pullrequestreview-580589728,

> Generally looks good, the only worry I could see would be a small edge case where if you add a mobject to a scene, then later add an interactive mobject to that mobject, it wouldn't be reflected in the scene.
> 
> But this is probably unusual enough it's not worth worrying about right now.

Also, The event handling mechanism was fragmented across Scene and Mobject classes.
The same has been abstracted and is contained within EventDispatcher class.

# Changes
1. Added manimlib/event_handler folder
> This contains all the logic behind event handling
2. manimlib/mobject/interactive.py, manimlib/mobject/mobject.py, manimlib/scene/scene.py
> To refactor event handling mechanism

# Result
ExampleWithEventDispatcher
The ControlsExample in example_scene works just as before, only the underlying structure is changed.


3b1b:
This looks like a much more sensible organization, thanks!

Issue #1330, opened 2021-02-01
Error after installing manim


AliNear:
I installed all manim & dependecies, but when I ran `python -m manim example_scenes.py OpeningManimExample`, I got the following error:
`Traceback (most recent call last):
  File "c:\users\jm\anaconda3\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\users\jm\anaconda3\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\jm\Documents\work\manim_new\manim\manim.py", line 5, in <module>
    manimlib.main()
  File "C:\Users\jm\Documents\work\manim_new\manim\manimlib\__init__.py", line 9, in main
    scenes = manimlib.extract_scene.main(config)
  File "C:\Users\jm\Documents\work\manim_new\manim\manimlib\extract_scene.py", line 113, in main
    scenes = get_scenes_to_render(all_scene_classes, scene_config, config)
  File "C:\Users\jm\Documents\work\manim_new\manim\manimlib\extract_scene.py", line 74, in get_scenes_to_render
    scene = scene_class(**scene_config)
  File "C:\Users\jm\Documents\work\manim_new\manim\manimlib\scene\scene.py", line 44, in __init__
    self.window = Window(self, **self.window_config)
  File "C:\Users\jm\Documents\work\manim_new\manim\manimlib\window.py", line 19, in __init__
    super().__init__(**kwargs)
  File "C:\Users\jm\Envs\manim.new\lib\site-packages\moderngl_window\context\pyglet\window.py", line 51, in __init__
    self._window = PygletWrapper(
  File "C:\Users\jm\Envs\manim.new\lib\site-packages\pyglet\window\win32\__init__.py", line 134, in __init__
    super(Win32Window, self).__init__(*args, **kwargs)
  File "C:\Users\jm\Envs\manim.new\lib\site-packages\pyglet\window\__init__.py", line 603, in __init__
    config = screen.get_best_config(config)
  File "C:\Users\jm\Envs\manim.new\lib\site-packages\pyglet\canvas\base.py", line 194, in get_best_config
    raise window.NoSuchConfigException()
pyglet.window.NoSuchConfigException`.
Any advice? And thank you

TonyCrane:
Try to comment out this line?
https://github.com/3b1b/manim/blob/fe85d4e02f6935c49fb0b88eebbd492dfff2d324/manimlib/window.py#L15

[...]

PR #1336, opened 2021-02-02
Package manimlib and automatically publish to pypi

TonyCrane:
I added the code to package manimlib and publish it on pypi using GitHub Action. Also fixed some bugs that made manimlib unable to be used normally after packaging.
- The first commit https://github.com/3b1b/manim/commit/9916f56fb10b8600856b37e1bb84b57f5cc63ad2: The package will not include `manim.py`, so the "manim" module will not be found. So I changed to get the manimlib directory, and returned to the previous level to get the original "manim_dir".
- The second commit https://github.com/3b1b/manim/commit/e4df135bbea9f6c8e57ef597e796bccd03b4fe58: Updated the version and document address in `setup.cfg`.
- The third commit https://github.com/3b1b/manim/commit/7a6a8556b87c3f7b307127dea100b8892dab0eba: The package will also not include `custom_default.yml`, so I created the same `default.yml` in the manimlib folder as the default configuration. At the same time, the previous `custom_default.yml` will take effect as before when the repo is cloned and run, because this personalized default value overrides the `default.yml` in manimlib
- The fourth commit https://github.com/3b1b/manim/commit/5b6997aea97de3d2f2ed697c5e60083644f85204: Set up a GitHub Action to package manimlib and publish it on pypi. Whenever a release is created, the current version will be released on pypi.
- The fifth commit https://github.com/3b1b/manim/commit/c5b00401663e15c8c34a450f4ce05c46cac1d3f9: As mentioned and solved in #1330, the line `samples=1` on some machines will cause `NoSuchConfigException`. So I deleted this line. But this may not be a good solution.
---
@3b1b, if you think it is okay to merge this PR, to package and publish manimlib to pypi, you also need to create two repository secrets in Settings-Secrets, one is `PYPI_USERNAME`, which is your pypi username, and the other is `PYPI_PASSWORD` , which is your pypi account password. After setting it up, create a release. If there is no error, the current version will be packaged and released to pypi at the same time. Then others can install manimlib through pip, and use manim directly as `manim code.py Scene` in anywhere.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,30 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+    
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
+        twine upload dist/*
--- a/README.md
+++ b/README.md
@@ -1,4 +1,8 @@
-![logo](logo/cropped.png)
+<p align="center">
+    <a href="https://github.com/3b1b/manim">
+        <img src="https://raw.githubusercontent.com/3b1b/manim/master/logo/cropped.png">
+    </a>
+</p>
 
 [![MIT License](https://img.shields.io/badge/license-MIT-blue.svg?style=flat)](http://choosealicense.com/licenses/mit/)
 [![Manim Subreddit](https://img.shields.io/reddit/subreddit-subscribers/manim.svg?color=ff4301&label=reddit)](https://www.reddit.com/r/manim/)
--- a/manimlib/config.py
+++ b/manimlib/config.py
@@ -152,7 +152,7 @@ def get_module(file_name):
 
 def get_custom_defaults():
     filename = "custom_defaults.yml"
-    manim_defaults_file = os.path.join(get_manim_dir(), filename)
+    manim_defaults_file = os.path.join(get_manim_dir(), "manimlib", "defaults.yml")
     with open(manim_defaults_file, "r") as file:
         custom_defaults = yaml.safe_load(file)
 
--- a/manimlib/config.py
+++ b/manimlib/config.py
@@ -134,7 +134,9 @@ def parse_cli():
 
 
 def get_manim_dir():
-    return os.path.dirname(inspect.getabsfile(importlib.import_module("manim")))
+    manimlib_module = importlib.import_module("manimlib")
+    manimlib_dir = os.path.dirname(inspect.getabsfile(manimlib_module))
+    return os.path.abspath(os.path.join(manimlib_dir, ".."))
 
 
 def get_module(file_name):
--- a/manimlib/defaults.yml
+++ b/manimlib/defaults.yml
@@ -1,0 +1,56 @@
+directories:
+  # Set this to true if you want the path to video files
+  # to match the directory structure of the path to the
+  # sourcecode generating that video
+  mirror_module_path: False
+  # Where should manim output video and image files?
+  output: ""
+  # If you want to use images, manim will look to these folders to find them
+  raster_images: ""
+  vector_images: ""
+  # If you want to use sounds, manim will look here to find it.
+  sounds: ""
+  # Manim often generates tex_files or other kinds of serialized data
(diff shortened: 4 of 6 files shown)


3b1b:
Before doing this, we should better check in with the manim community to think about what the two versions of manim look like on pypi.  Let me get back to you here.

sahilmakhijani:
@TonyCrane, before publishing this to pypi, please take a look at https://github.com/3b1b/manim/pull/1338

TonyCrane:
> Before doing this, we should better check in with the manim community to think about what the two versions of manim look like on pypi. Let me get back to you here.

Their version is called [manim](https://pypi.org/project/manim/) on pypi, and there is no conflict in installation.
But both of them and our console scripts include `manim`, which may cause conflicts after the two versions are installed at the same time.

3b1b:
I think we should have a name other than "manimlib" that better references the distinction between the two.  Would it be strange to name it "manim3b1b"?  I would say manimgl, but it sounds like the community has plans to adapt the OpenGL backend here for that version too.  What are your thoughts?

TonyCrane:
I think it is strange to name it "manim3b1b". It is not a good idea to include your channel name in the name of the package, which may make people think that the package is only used by 3b1b himself.
In my opinion, it might be a good idea to name it "manimgl" or "manimGL". What ManimCommunity is doing should support multiple rendering backends (including cairo, WebGL, OpenGL), so maybe this is not conflicting, and it can also reflect the special features of 3b1b/manim.

3b1b:
Right, at least for where they both are now, manimgl would most accurately capture part of the distinction.  And even when they do adapt the OpenGL backend here to their repo, it would at least convey the less flexible backend here :)

What's involved in changing the name on pypi?

TonyCrane:
Maybe we need to rename the manimlib folder to manimgl, but this will cause the history of the source code to be lost.
Then just change the name in setup.cfg to change the name to manimgl

TonyCrane:
So I still doubt whether we need to change the package name from manimlib to manimgl. The two versions do not conflict here (we use manimlib, they use manim). But we have conflicts on the entry point (we both are manim), maybe we only need to change console_scripts to `manimgl = manimlib:main` to run through `manimgl code.py Scene`.

3b1b:
Does git have no way to rename a folder while preserving its history?

It's not super necessary to change names; I'll go ahead and just merge this and get things going.  It's just that manimlib doesn't really feel descriptive.

3b1b:
So with that, shall I go ahead and make a new release?

TonyCrane:
You should set `PYPI_USERNAME` and `PYPI_PASSWORD` in secrets first.

3b1b:
I believe that's already done.

Any objection to labeling this release 1.0.0?

TonyCrane:
Before publishing it, I think we should rename the entry point to `manimgl`. What about you?
It is certainly no problem to label it as 1.0.0

3b1b:
It seems a little messy with the repo being named manim, the pypi presence being under manimlib, and an entry point named manimgl.

If we're going with manimlib on pypi, then for consistency we should also make the entry point manimlib.  Thoughts?

TonyCrane:
However, the name manimlib is more like a library, and using it as an entry point does not seem appropriate.
Perhaps names like `manim-run` , `manim-render`, etc. sound good as entries?

3b1b:
Right, I see what you are saying.  Is there a reason not to do "manimgl" as both the entry point and the pypi name, but then leave the manimlib folder named as it is?  Sorry if that's a dumb question, but it sounded like maybe there was a reason to change the manimlib folder name if we also named it something different on pypi.

TonyCrane:
It would be terrible to rename both the folder and pypi. Git cannot migrate the history of folders, so the historical versions and change records of these files cannot be viewed easily on GitHub. This is also very inconvenient.
Because there is no conflict between us and the community in this respect, I still recommend keeping the original "manimlib" folder name and pypi name. Although this does not clearly highlight the characteristics of 3b1b/manim, at least it will not cause more trouble.
But on entry point, we and community both use "manim", so I think we'd better change the entry point name to a new one, such as "manimgl", "manim-render", "manim-run", etc.
In this way, we can still install 3b1b/manim via `pip install manimlib`, import via `from manimlib.imports import *`, and then run it via `manimlib(or manim-render, etc.) code.py Scene`.

3b1b:
I just answered this in email but copying here for the record.  I was only asking if there's anything explicitly wrong with this configuration:

- Pip name: manimgl
- Entry point: manimgl
- Folder name (unchanged): manimlib

I would be alright with what you're suggesting, which is certainly easiest and probably the best:

- Pip name: manimlib
- Entry point: manim-render
- Folder name (unchanged): manimlib

We could also add a note that for anyone who doesn't plan to also have the community edition installed, suggesting that they alias "manim" to "manim-render".

TonyCrane:
Sorry I may have misunderstood you.
This is of course no problem, but still uses `from manimlib.imports import *` after installing manimgl from pypi, which may be a bit strange. 
But this can also distinguish the current version from cairo-backend.
The naming is also good, so I just changed it in https://github.com/3b1b/manim/commit/b558ae98cf4ad1f6af13a680e820001c92ad63e6.
For the entry point, I set two names manimgl and manim-render.

PR #1345, opened 2021-02-03
Converted 0 to 0.0 in sdf() and angle_between_vectors()

samipism:
## Motivation

GLSL files viz; geom.glsl and frag.glsl has two functions angle_between_vectors and sdf. Their return type is float but returned 0(int) which caused glsl error in many gpus.

## Proposed changes

- Converted the 0 to 0.0 in return statement of angle_between_vectors and sdf


TonyCrane:
The same as #1329, which has already been merged, but was reverted by me wrongly.

Issue #1243, opened 2020-10-01
Manim current status and contribution guidelines


Tony031218:
Hello everyone, I will explain the current status of manim in this issue.

Now there are _three_ main manim versions, and their differences:
1. **The master branch of [3b1b/manim](https://github.com/3b1b/manim)**: (probably the most used currently) rendering on the CPU using cairo, currently stable.
2. **The shaders branch of [3b1b/manim](https://github.com/3b1b/manim)**: (Grant is making, using and perfecting) rendering on GPU using modernGL.
3. **[ManimCommunity/manim](https://github.com/ManimCommunity/manim)**: (@ManimCommunity is being improved and will be released on pypi) also uses cairo to render on the CPU, but has changed a lot of structure and added many new features.

If you want to contribute to manim:
- **The master branch of [3b1b/manim](https://github.com/3b1b/manim)** currently only accepts pull requests that fixes bugs / fixes typos / improves existing content, and it is _not recommended_ to add new features to it.
- **[ManimCommunity/manim](https://github.com/ManimCommunity/manim)** accepts a wider range of new features and pull requests.
- **[manim-kindergarten/manim_sandbox](https://github.com/manim-kindergarten/manim_sandbox)** has a lot of new common classes in the `utils/` folder, and accepts pull requests.

Lastly, thank you very much for using manim and willing to contribute. Enjoy yourself ! 😄

qo4on:
What is the current status of the shaders branch?

Tony031218:
> What is the current status of the shaders branch?

3b1b is making and perfecting the manim in the shaders branch, which will be faster.  But there may be some problems and not stable at present.  If you have the ability to solve the problems you met, you can try to use it.  But for the pull requests for shaders branch, 3b1b may handle it by himself, in other words, it will make you wait for a long time and may not be merged.

ShrirangPatil:
In Android, their is no way of making animation or graphs easily, can we use this library to make graphs and animations in android apps??

iliya-malecki:
> In Android, their is no way of making animation or graphs easily, can we use this library to make graphs and animations in android apps??

[...]

PR #1357, opened 2021-02-05
handle_play_like_call decorator to use functools.wraps

parmentelat:
this is a remake of https://github.com/3b1b/manim/pull/1204 on the current master

this way, decorated methods - typically Scene.play - have a decent docstring
so first-time users can use help()

### DIFF
--- a/manimlib/mobject/mobject.py
+++ b/manimlib/mobject/mobject.py
@@ -3,6 +3,7 @@ import itertools as it
 import random
 import sys
 import moderngl
+from functools import wraps
 
 import numpy as np
 
@@ -1287,6 +1288,7 @@ class Mobject(object):
     # Operations touching shader uniforms
 
     def affects_shader_info_id(func):
+        @wraps(func)
         def wrapper(self):
             for mob in self.get_family():
                 func(mob)
--- a/manimlib/mobject/types/vectorized_mobject.py
+++ b/manimlib/mobject/types/vectorized_mobject.py
@@ -2,7 +2,7 @@ import itertools as it
 import operator as op
 import moderngl
 
-from functools import reduce
+from functools import reduce, wraps
 
 from manimlib.constants import *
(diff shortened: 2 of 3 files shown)


TonyCrane (approved):
But there are two more wrappers. You can also improve them by the way:
https://github.com/3b1b/manim/blob/8f52c7bda8e298d15d3958ba76bc56648d4a9e19/manimlib/mobject/mobject.py#L1289-L1295 https://github.com/3b1b/manim/blob/8f52c7bda8e298d15d3958ba76bc56648d4a9e19/manimlib/mobject/types/vectorized_mobject.py#L809-L816

parmentelat:
that's right, I had proposed to take care of that in my initial PR, but had forgotten about all this in the meanwhile :)

I am in no position to test all  this though at this time, the code I had written in summer last year looks broken with the current version (where is the `Paragraph` class now ?), and I could not spot any automated tests (did I miss them ?)

in any case, here you go ...

TonyCrane:
This version has been developed since last year, so `Paragraph` and `Code` are not included in it. For testing, 3b1b/manim probably has never had a well-formed and complete test. But ManimCommunity/manim has.

parmentelat:
are you suggesting I use the community version instead if I need `Paragraph` and `Code` ? otherwise where shoud I grab them ?
more generally, I was not aware of these 2 flavours, where should I read about that do get up to speed ?

many thanks for the feedback !

TonyCrane:
It may be that `Paragraph` and `Code` were mistakenly missed when merging shaders branch into master, but you can still use the old version that contains them in the `cairo-backend` branch now. 
For ManimCommunity/manim, there is a big difference between the two versions at present, and there is a brief description of the difference in #1243. These two versions all have advantages, which one you use depends on you.
If you want to talk more, maybe you can start a discussion instead of this pr.

PR #1379, opened 2021-02-10
Add .animate syntax

friedkeenan:
## Motivation

Adding `.animate` syntax greatly improves the syntax for animating mobject method calls, especially with regards to passing keyword arguments to methods. Where before you would do

```py
self.play(mobj.method, arg, {"kwarg": value})
```

Now you can do

```py
self.play(mobj.animate.method(arg, kwarg=value))
```

(Though the old syntax still works).

For more information, see the [Manim Community docs](https://docs.manim.community/en/stable/reference/manim.mobject.mobject.Mobject.html#manim.mobject.mobject.Mobject.animate).

## Proposed changes

- Add the `animate` property to the `Mobject` class, along with all associated logic (`override_animate`, using the `.animate` syntax in `AnimationGroup`s, etc.).
- Update all usages of the old syntax to the `.animate` syntax, besides from some lines in the docs/examples explaining the old syntax.
- My text editor is configured so that it removes trailing white space and adds a newline at the end of every file I save, that seems to have affected a few places.

## Test

**Code**:
```py

class Test(Scene):
    def construct(self):
        c = Circle()
        self.add(c)
        self.play(c.animate.scale(2))
        self.play(c.scale, 0.5)
```
**Result**:

https://user-images.githubusercontent.com/20881398/107535195-237b3980-6b86-11eb-99ac-bb4783a6045c.mp4


3b1b:
Looks great, thanks for bringing it over!

PR #1381, opened 2021-02-11
Allow setting scene_config via Scene.CONFIG

abhi-6988:
## Motivation

In the `cairo-backend` branch, configurations such as `background_color` of `Camera` was possible to be set via `Scene`'s `CONFIG` like so:
```py
Class MyScene(Scene):
    CONFIG = {"camera_config": {"background_color": <color>}}
```
But it is currently not possible in the `master` branch. The only way to achieve this is to either change the `custom_defaults` or pass the relevant command line flags (for example `-c <color>` for background color). The reason is, `scene_config` built through `custom_defaults` and the command line flags no where considers `MyScene.CONFIG`. The `digest_config` function might seem to take care of this, but since `kwargs` are given higher priority in `digest_config` and `scene_config` is indeed passed as `kwargs` while instantiating the `Scene`,https://github.com/3b1b/manim/blob/787b3be357e3c8b6c119aa45f1b8101fdee70487/manimlib/extract_scene.py#L76
`MyScene.CONFIG` gets overridden by the `scene_config` while the opposite is expected. 
## Proposed changes

- Allow users to set `camera_config`, `window_config`, `file_writer_config` and other configurations of a `Scene` through `CONFIG` dict of their scenes. 

## Test

**Code**:
```py
class Test(Scene):
    CONFIG = {
        "camera_config":{
            "background_color": "#cc9900",
            "samples": 4
        },
        "window_config":{"size": (1900, 1000)},  # works if preview is True
        "preview": False,
        "file_writer_config": {
            "save_last_frame": True,
            "open_file_upon_completion": True,
        }
    }

    def construct(self):
        motivation = Text(
            "Allow setting scene_configs through CONFIG dict", 
            font="Georgia",
            t2c={"scene_config": BLACK, "CONFIG": BLACK}
        ).set_width(FRAME_WIDTH-1)
        self.add(motivation)
```

**Result**:
- **command** : `python -m manim .\test.py Test`
- **Result on master**: (All the configs are ignored, a window at top right corner opens up which is the default behavior)
Screenshot (43)
- **Result on current branch**: (All the configs are respected and the Test.png file opens up as expected from `Test.CONFIG`)
Test


manimlib/extract_scene.py
```diff
@@ -73,6 +74,11 @@ def get_scenes_to_render(scene_classes, scene_config, config):
         found = False
         for scene_class in scene_classes:
             if scene_class.__name__ == scene_name:
+                if hasattr(scene_class, "CONFIG"):
+                    scene_class_config = scene_class.CONFIG
+                else:
+                    scene_class_config = {}
+                scene_config = merge_dicts_recursively(scene_config, scene_class_config)
```
TonyCrane, line 81: This will override the CLI flags like `-c` (`--color`)

TonyCrane (changes requested):
Just making these changes will cause many problems. For example, this will completely override the configuration passed in through CLI flags.
And pay attention to lines 69~70:
https://github.com/3b1b/manim/blob/ae371538c0e9d533dee31d3a8c5679866ec7032d/manimlib/extract_scene.py#L69-L70
 if there is only one scene in this file and `python manim.py code.py` is used (the scene to be rendered is not specified), or use `python manim.py code.py -aw` to export all scenes, `Scene.CONFIG` will still not be used.

TonyCrane:
I discovered this problem early, and I also thought about its solution.
Here I plan to straighten out my thinking. Taking the control of the background color of the scene as an example, we have three methods:

1. Modify `style.background_color` in `custom_defaults.yml`
2. Write `Scene.CONFIG`, add `camera_config` (including `background_color`)
3. Use CLI flags directly. Such as `python manim.py code.py -c BLACK`

Among them, 

1. is the default value for all scenes
2. is the value set for a single scene
3. is the value specified for a single run.

So, looking at it this way, from the user's point of view, the reasonable priority order should be 
```
CLI flags > Scene.CONFIG > custom_defaults.yml (3>2>1)
```
However, the priority before this pr is `CLI flags > custom_defaults.yml > Scene.CONFIG (3>1>2)`, which completely override `Scene.CONFIG`.
The priority of your pr is `Scene.CONFIG > CLI flags > custom_defaults.yml (2>3>1)`, which completely override CLI flags.

The difficulty of fixing it is: https://github.com/3b1b/manim/blob/787b3be357e3c8b6c119aa45f1b8101fdee70487/manimlib/config.py#L274-L276
Here, CLI flags and `custom_defaults.yml` are considered together, so `Scene.CONFIG` is difficult to insert in the middle (especially there are more than one scene to render), unless we build `scene_config` in the correct order again elsewhere.

manimlib/extract_scene.py
abhi-6988: > This will override the CLI flags like `-c` (`--color`)

That should be the intended behavior though? I mean it wouldn't make sense to use both `CONFIG` and CLI flag to change the background color of a given scene. One or the other has to be overridden. And as you mentioned below, this PR goes with the priority order `Scene.CONFIG > CLI flags`

abhi-6988:
> if there is only one scene in this file and `python manim.py code.py` is used (the scene to be rendered is not specified), or use `python manim.py code.py -aw` to export all scenes, `Scene.CONFIG` will still not be used.
Yeah, that's true thanks for pointing out. I'll try to fix this

manimlib/extract_scene.py
TonyCrane: Yes, we rarely use this, but it does not mean that we should not support such use.
Sometimes, you may try to change some other colors to render different results. At this time, it is very convenient to use CLI flags, but if you happen to set the color in `CONFIG` at this time, it will completely fix the background color.
Therefore, if there is a conflict between CLI flags and `CONFIG`, the CLI flags which is closer to the call should be kept.
In addition, the old version in `cairo-backend` supports this (3>2>1 as I said below)

TonyCrane:
Using CLI flags is the most convenient method when debugging, the purpose is to debug by yourself, temporarily specify the background color, rather than permanently save where for others to use. 
So we shouldn't completely cover its function, right?
In fact, the 3. I mentioned is not commonly used, so it seems that my suggestion and your changes are both 2>1, which is no problem. But if we can put 3. which is the closest to the user side in the first place, it will be much better.

manimlib/extract_scene.py
abhi-6988: Yeah, got it. I didn't know that the `cairo-backend` supported 3>2>1. I'll try to get this order working with this PR, thanks for the review!

abhi-6988:
> Using CLI flags is the most convenient method when debugging, the purpose is to debug by yourself, temporarily specify the background color, rather than permanently save where for others to use.
> So we shouldn't completely cover its function, right?
> In fact, the 3. I mentioned is not commonly used, so it seems that my suggestion and your changes are both 2>1, which is no problem. But if we can put 3. which is the closest to the user side in the first place, it will be much better.

Yeah, that makes sense. I'll try to get the 3>2 priority working.

abhi-6988:
Hi @TonyCrane , I was thinking about how to implement the priority order. To me, the only way to get this done seems to be creating a new `scene_cli_config` dict whose key-value pairs are the arguments and their corresponding values passed explicitly  through cli. And then the ultimate `config` that will be passed in the `Scene`s instantiation would be the result of `merge_dicts_recursively(scene_config, scene_class_config, scene_cli_config)`.

The `scene_cli_config` will be the result of the following function:
```py
def get_scene_cli_config(args):
    # depends on the fact that all the args are currently either None or False
    passed_args = {}
    for key, value in vars(args).items():
        if value:
            passed_args[key] = value
    return passed_args
    
scene_cli_config = get_scene_cli_config(args)
``` 
Depending on the keys of `scene_cli_config`, it has to be sorted out into `camera_config`, `window_config` etc and then the final config would be `merge_dicts_recursively(scene_config, scene_class_config, scene_cli_config)`.
Would this be a viable solution? I couldn't think of any other way.

TonyCrane:
This is indeed viable, but at present, `get_configuration(args)` is based more on `args`. After getting `config` from `args` through `get_configuration(args)`, then obtaining `scene_cli_config` from `args` through `get_scene_cli_config(args)` again, there may be some duplication.

Think about it again. What is the reason why the new version cannot use the priority of "3>2>1" like the old version:

- There is no `custom_config.yml` in the old version. The `get_configuration(args)` function only processes `config` based on `args`. The default value is specified by the `CONFIG` of the parent class (`Scene`, `SceneFile`, `Camera`, etc.) in the source code. This will be covered by the `CONFIG` of the sub scene class that needs to be rendered, so "2>1" is achieved. Finally, the `config` obtained according to `args` is passed into the `Scene` as `kwargs`. According to the priority specified in `digest_config()`, `kwargs` will overwrite `CONFIG`, which will achieve "3>2".
- But there is `custom_config.yml` in the new version, and both `args` and `custom_config` are considered in the `get_configuration(args)` function at the same times, which guarantees "3>1". But the `config` returned by `get_configuration()` was passed into the `Scene` as `kwargs`, overwriting `CONFIG`, which resulted in "(3>1)>2".

So considering the difference between the new and the old version, maybe we can ignore `custom_config` in `get_configuration(args)` and only consider `args`. Then extract the content related to `scene_config` in `custom_config` separately, pass it into `Scene` separately, and convert it into attributes before calling `digest_config()` (this ensures that `CONFIG` will overwrite `custom_config`). In this way, the `config` obtained by `args` can overwrite `CONFIG`, and the `CONFIG` can overwrite `custom_config`, which ensure "3>2>1".

This is my idea.
btw, @3b1b, what do you think of this problem?

abhi-6988:
> maybe we can ignore `custom_config` in `get_configuration(args)` and only consider `args`. Then extract the content related to `scene_config` in `custom_config` separately, pass it into `Scene` separately, and convert it into attributes before calling `digest_config()` (this ensures that `CONFIG` will overwrite `custom_config`)

That's a good idea, thanks I'll work in this direction :)

PR #1383, opened 2021-02-12
Use ManimPango for Text rendering

naveen521kk:
## Motivation
Pango is a lot better at Text Rendering rather than using a Toy API for Cairo. I had discussed this with @3b1b on [Discord.](https://discord.com/channels/581738731934056449/712140817200054294/808055730639863868). This will remove the dependency on pycairo.

It also provides another method to add a font file to Pango search path, whose family name can later be called. This method isn't available for macOS yet.

I couldn't find `Code` here, that would require a small change setting `disable_ligatures`, or else colours would be broken, due to ligatures.

## Proposed changes
- requirements.txt - add manimpango
- text_mobject.py - use manimpango

## Test
Running example scenes should work, no breaking API changes. `register_font` API can be used as
```py
with register_font('roboto.ttf'):
    a=Text("Hello Roboto",font="Roboto")
```
A list of fonts visible to Pango can be got from
```py
>>> import manimpango
>>> manimpango.list_fonts()
[ ... ]
```
More API at https://manimpango.manim.community/en/latest/

### DIFF
--- a/manimlib/mobject/svg/text_mobject.py
+++ b/manimlib/mobject/svg/text_mobject.py
@@ -1,27 +1,21 @@
-import re
-import os
 import copy
 import hashlib
-import cairo
+import os
+import re
+import typing
+from contextlib import contextmanager
+from pathlib import Path
+
+import manimpango
 from manimlib.constants import *
 from manimlib.mobject.geometry import Dot
 from manimlib.mobject.svg.svg_mobject import SVGMobject
 from manimlib.utils.config_ops import digest_config
 from manimlib.utils.customization import get_customization
-from manimlib.utils.directories import get_text_dir
-
+from manimlib.utils.directories import get_downloads_dir, get_text_dir
+from manimpango import PangoUtils, TextSetting
 
 TEXT_MOB_SCALE_FACTOR = 0.001048
-
-
-class TextSetting(object):
-    def __init__(self, start, end, font, slant, weight, line_num=-1):
-        self.start = start
-        self.end = end
-        self.font = font
-        self.slant = slant
-        self.weight = weight
-        self.line_num = line_num
 
 
 class Text(SVGMobject):
@@ -44,6 +38,7 @@ class Text(SVGMobject):
         "t2g": {},
         "t2s": {},
         "t2w": {},
+        "disable_ligatures": True,
     }
 
     def __init__(self, text, **config):
@@ -55,12 +50,11 @@ class Text(SVGMobject):
             text_without_tabs = text.replace('\t', ' ' * self.tab_width)
         self.text = text_without_tabs
         file_name = self.text2svg()
-        self.remove_last_M(file_name)
-        self.remove_empty_path(file_name)
+        PangoUtils.remove_last_M(file_name)
         SVGMobject.__init__(self, file_name, **config)
         self.text = text
-        self.apply_space_chars()
-
+        if self.disable_ligatures:
+            self.apply_space_chars()
         if self.t2c:
             self.set_color_by_t2c()
         if self.gradient:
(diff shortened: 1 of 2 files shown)


TonyCrane (approved):

3b1b:
Is this the intended behavior?

image

```
class TextExample(Scene):
    def construct(self):
        fonts = Text(
            "And you can also set the font according to different words",
            font="Arial",
            t2f={"font": "Consolas", "words": "Consolas"},
            t2c={"font": BLUE, "words": GREEN}
        )
        fonts.set_width(FRAME_WIDTH - 1)
        self.play(Write(fonts))
        self.wait()
```

naveen521kk:
Looks like a bug to me. I tried doing the same using Community edition's MarkupText (I didn't bring it here because CONFIG was something out of my understanding), and also the legacy CairoText. I get a video like
```py
from manim import *
class Font(Scene):
    def construct(self):
        fonts = Text(
            "And you can also set the font according to different words",
            font="Arial",
            t2f={"font": "Consolas", "words": "Consolas"},
            t2c={"font": BLUE, "words": GREEN},
            disable_ligatures=True
        )
        fonts_c = CairoText(
            "And you can also set the font according to different words",
            font="Arial",
            t2f={"font": "Consolas", "words": "Consolas"},
            t2c={"font": BLUE, "words": GREEN},
        )
        markup = MarkupText(
            'And you can also set the <color col="BLUE"><span font_desc="Consolas">font</span></color> according to different <color col="GREEN"><span font_desc="Consolas" color="GREEN">words </span></color>',
            disable_ligature = True,
            font="Arial",
        )
        fonts.scale(.8)
        self.play(Write(fonts))
        self.wait()
        self.clear()
        fonts_c.scale(.8)
        self.play(Write(fonts_c))
        self.wait()
        self.clear()
        markup.scale(.6)
        self.play(Write(markup))
        self.wait()
```

https://user-images.githubusercontent.com/49693820/107844398-11ed9800-6df9-11eb-82f2-23041211068b.mp4

so it seems like a bug in aligning texts in ManimPango for `Text`. Can you open an issue in that [repo](https://github.com/manimcommunity/manimpango)?

TonyCrane:
> Is this the intended behavior?
> 
> image
> 
> ```
> class TextExample(Scene):
>     def construct(self):
>         fonts = Text(
>             "And you can also set the font according to different words",
>             font="Arial",
>             t2f={"font": "Consolas", "words": "Consolas"},
>             t2c={"font": BLUE, "words": GREEN}
>         )
>         fonts.set_width(FRAME_WIDTH - 1)
>         self.play(Write(fonts))
>         self.wait()
> ```

image
But why I don't have this problem?

naveen521kk:
@TonyCrane I think you don't have the font installed or something like that? Btw which OS are you using?

TonyCrane:
>  I think you don't have the font installed or something like that? Btw which OS are you using?

But the result fonts are correct ("font" and "words" are Consolas, the rest are Arial), so I made sure I have installed the fonts.
I am using Windows 10, Python 3.7.3, ManimPango 0.2.2, and the manim source code is the latest.

naveen521kk:
I think maybe caching issue? Did you run the same example before updating to recent branch which uses pango and the svg file from cairo is cached? That's should be case I think. But still not sure...

TonyCrane:
Oh, sorry, I just cleaned the cache of Text, and got the same result as you.
image
But if I kept the `disable_ligatures` as default `True`, then the coloring will be incorrect:
DYY} }`BQ6(Z} 0R7@_}HYQ

naveen521kk:
I don't think that will fix this. The colouring is a bit off because of ligatures where in two letter comes in a single path element of the SVG file. Pango would handle it but the svg parser in both the version isn't that great to handle styles. I have some plans to make it not use SVG instead directly return each character as an ImageMobject or rather return a set of points, but that requires work and isn't yet done. Once that's done this problem will be fixed 😉.

eulertour:
This PR causes a crash immediately for me and it looks like it's causing other problems from the earlier posts. Can this at least be made optional until it can be done stably?

naveen521kk:
> This PR causes a crash immediately for me

There could be much reason why things crash. ManimPango is necessary for Community Version also. If you say this crashes for you, it should crash for ManimCE also...

Issue #1416, opened 2021-02-25
Error using Elbow()


yerovi84:
### Describe the error
I tried to use the Elbow() object but I got an error.
 
### Code and Error
**Code**:
class ElbowExample(Scene):
    def construct(self):
        triangle = Polygon(ORIGIN, RIGHT, RIGHT+UP)
        elbow = Elbow(color=RED)
        elbow.set_points_as_corners([ORIGIN, RIGHT, RIGHT+UP])
        elbow.set_width(elbow.width, about_point=RIGHT+np.array([-0.1, 0.1, 0.0]))
        self.add(triangle)
        self.add(elbow)
        self.wait(2)

**Error**:
...geometry.py", line 572, in __init__ 
in __init__
    super().__init__(self, **kwargs)
TypeError: __init__() takes 1 positional argument but 2 were given

### Environment
**OS System**: Win10 
**manim version**: version of 23/02/2021
**python version**: 3.8.1

JWro:
The problem is passing `self` to `super().__init__`, see https://github.com/3b1b/manim/pull/1419/commits/63b497c352eff38cd5b16b32fd9f385fd902a59f

PR #1419, opened 2021-02-26
Fix init of Elbow super class

JWro:
## Motivation
Elbow is unusable in the master

## Proposed changes
Fixed the __init__ of Elbow

## Test
**Code**:
class ElbowExample(Scene):
    def construct(self):
        elbow = Elbow()
        self.add(elbow)
**Result**:
An Elbow is drawn


JWro:
This is connected to issue https://github.com/3b1b/manim/issues/1416

Issue #1391, opened 2021-02-16
Nothing happens when manimgl is run!


heethesh:
### Describe the error
No output on the console or no GUI shows up when `manimgl` is run from the command line (installed by `pip install manimgl`). I observe the same issue when running from this repository and installing using `pip install -e .` as well.

### Code and Error
```
manimgl
```

### Environment
**OS System**: Ubuntu 20.04.1 (5.8.0.38-generic)
**OpenGL version**: 4.6.0 NVIDIA 450.102.04
**manim version**: 1.0.0 (`pip install manimgl`)
**python version**: Tried both 3.6 and 3.8

WiredPhysics:
You need to pass some flag(s) for it to work Try `manimgl example_scenes.py OpeningManimExample` or `manimgl example_scenes.py OpeningManimExample -w` to write the scene to a file

gnap114039:
same issue here !

OS System: Ubuntu 20.04.1 (5.8.0.43-generic)
PyOpenGL: 3.1.5
manim version: 1.0.0 (pip install manimgl)
python version: 3.8.5

I have tried the following command line:
> manimgl example_scenes.py TextExample
> manimgl example_scenes.py TextExample -w
> manim-render example_scenes.py OpeningManimExample

the terminal will wait command return after hit enter, and htop showing CPU single thread almost 100 % usage. 



> ### Describe the error
> No output on the console or no GUI shows up when `manimgl` is run from the command line (installed by `pip install manimgl`). I observe the same issue when running from this repository and installing using `pip install -e .` as well.
> 
> ### Code and Error
> ```
> manimgl
> ```
> 
> ### Environment
> **OS System**: Ubuntu 20.04.1 (5.8.0.38-generic)
> **OpenGL version**: 4.6.0 NVIDIA 450.102.04
> **manim version**: 1.0.0 (`pip install manimgl`)
> **python version**: Tried both 3.6 and 3.8

gnap114039:
same issue here !

OS System: Ubuntu 20.04.1 (5.8.0.43-generic)
PyOpenGL: 3.1.5
manim version: 1.0.0 (pip install manimgl)
python version: 3.8.5

I have tried the following command line:
> manimgl example_scenes.py TextExample
> manimgl example_scenes.py TextExample -w
> manim-render example_scenes.py OpeningManimExample

the terminal will wait command return after hit enter, and htop showing CPU single thread almost 100 % usage.

gnap114039:
same issue here !

[...]

Issue #1387, opened 2021-02-14
Segfault when running manimgl


eulertour:
### Describe the bug
Running `manimgl` with or without arguments results in a segfault

**Code**:
n/a

**Wrong display or Error traceback**:
![printscreen](https://user-images.githubusercontent.com/43117506/107866198-e7183980-6e22-11eb-8afc-452a65efc4fc.png)

### Additional context
The crash began after #1383 was merged.

naveen521kk:
@eulertour Can you run with the debugger and post the backtrace?
```sh
gdb -batch -ex "run" -ex "bt" manimgl
```

naveen521kk:
Also, if the problem is with manimpango, can you check whether it works with ManimCE?

TonyCrane:
I went back to the version you mentioned, and then created a new conda environment, installed manimgl and ran it without any errors.
```sh
git checkout ad332c586341a665b2e69d4dbea8b023c66d575d
conda create -n manim_1387 python=3.8
source activate manim_1387
pip install -e . -i https://pypi.tuna.tsinghua.edu.cn/simple/
manimgl
```
And I feel that there is nothing in #1383 that will cause a segfault

eulertour:
@naveen521kk The first command
![printscreen](https://user-images.githubusercontent.com/43117506/107867767-5b0e0e00-6e32-11eb-8dba-896a1af2c542.png)

the second command
![printscreen](https://user-images.githubusercontent.com/43117506/107867779-7da02700-6e32-11eb-8b7f-d3521dec248c.png)

it works with ManimCE.

@TonyCrane If you weren't aware, ManimPango is a custom-written C extension for using Pango.
https://github.com/ManimCommunity/ManimPango/blob/main/manimpango/cmanimpango.pxd

naveen521kk:
does this command show Seg Fault?
```sh
python -c "import manimpango"
```

eulertour:
No

naveen521kk:
for the results about using gdb, you don't have debugging symbols installed see https://wiki.python.org/moin/DebuggingWithGdb

eulertour:
You don't have to help, I almost never use `Text` anyway.

I'm suggesting to make the change optional because the functionality isn't necessary to use manim and the code is unstable.

naveen521kk:
rather try running python using the verbose mode, 
```sh
python -v -m manimlib
```

naveen521kk:
[...]

PR #1421, opened 2021-02-26
update docs for linux installation

naveen521kk:
## Motivation
See 
https://github.com/3b1b/manim/issues/1387
https://github.com/3b1b/manim/issues/1391

ManimPango should be installed from the source for Linux users.
Also, see https://github.com/ManimCommunity/ManimPango/issues/27

## Proposed changes
update docs


3b1b:
Thanks!

PR #1530, opened 2021-06-05
Modified the shaders slightly

Wallbreaker5th:
## Motivation
- `camera_distance` is set to be 6 and doesn't change with the `focal_distance`.
- The z coordinate is used to check whether the surface is facing the camera. It generates the wrong result which is particularly obvious when `focal_distance` is small.

## Proposed changes
- Replace `float camera_distance = 6;` with `float camera_distance = focal_distance;`, and add `uniform float focal_distance;` in some shaders.
- Use the dot product to check whether the surface is facing the camera.

## Test
**Code**:
```python
from manimlib import *


class Test(Scene):
    CONFIG = {
        "camera_class": ThreeDCamera,
        "camera_config": {
            "focal_distance": 0.5,
            "light_source_position": [1, 1, -2],
            "frame_config":{
                "frame_shape": (4, 3),
            }
        }
    }

    def construct(self):
        sphere = Sphere(radius=2)
        sphere.shift(DOWN*3.3)
        sphere.set_color('#cccccc')
        self.add(sphere)
        self.wait(1)

```

**Result**:
Before:
image

After:
image

### DIFF
--- a/manimlib/shaders/inserts/finalize_color.glsl
+++ b/manimlib/shaders/inserts/finalize_color.glsl
@@ -17,8 +17,7 @@ vec4 add_light(vec4 color,
                float shadow){
     if(gloss == 0.0 && shadow == 0.0) return color;
 
-    // TODO, read this in as a uniform?
-    float camera_distance = 6;  
+    float camera_distance = focal_distance;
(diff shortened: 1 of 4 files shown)


3b1b:
Thanks!

Issue #1405, opened 2021-02-23
Weird line drawing and improper text sizing using manim's example_scenes.py


mrmcsoftware:
![ex1](https://user-images.githubusercontent.com/28698130/108806192-032c8100-756f-11eb-86e2-62d4b6cc3981.png)
![ex2](https://user-images.githubusercontent.com/28698130/108806207-0c1d5280-756f-11eb-87b3-ee5d956047b8.png)
![ex3](https://user-images.githubusercontent.com/28698130/108806216-13dcf700-756f-11eb-923b-7dec55748ce3.png)
![ex4](https://user-images.githubusercontent.com/28698130/108806222-18091480-756f-11eb-9360-4bc7c373e718.png)

### Describe the bug
The first two images show a weird line drawing problem I'm having while running manim's supplied example_scenes.py file.  Instead of drawing lines, it draws triangles.  I can't imagine that's intended, but let me know if it is.  This problem happens in the other example scenes.

The last two images show another problem - the text doesn't fit on the screen correctly.  I can change the font to something that will fit, but I probably shouldn't have to.  Plus, I imagine the grid should be showing the complete range of -10 to 10 instead of only part of it.  So, I guess it's really an image scaling problem and not a text problem.  Changing resolution doesn't solve it.

Also, ControlsExample doesn't seem to do anything. 

**Code**:
Just refer to example_scenes.py

### Additional context
I'm using Python 3.8 on Ubuntu 20.10

mrmcsoftware:
I tried some other scene (from one of the reported issues that shows what it's supposed to look like) and the results are even worse.  BTW, adding to my issue report, to be more precise, my version of Python is 3.8.6

![ex5](https://user-images.githubusercontent.com/28698130/108885279-9c8e7e00-75d5-11eb-8c11-9510291ca189.png)

acnq:
I am also working around installing the master branch of manim, but my Problem seems more serious
Firstly, the mis-size of the text still exist in my version
![USUN1)0GKNC7 K{1}EQAL1V](https://user-images.githubusercontent.com/62015811/109169164-7d047a80-77ba-11eb-9e5f-9aee58244541.png)
secondly, a lot of grid in OpeningManimExample is missing, 
![GK~O{~O4Q1LHYR)6`)@}%E8](https://user-images.githubusercontent.com/62015811/109169476-cfde3200-77ba-11eb-9f3d-054fa6e45a19.png)
so does the coordinates in the next Animation
[...]

Issue #1399, opened 2021-02-20
Problems with Manimgl


andymeb:
### Describe the error

I'm getting weird graphics when I run the examples_scene and also many warnings. I have installed all miktex packages and deleted the \usepackage[T1]{fontenc} but I'm still not able to make it work. I'm not sure what could be wrong at this point. 
### Code and Error
**Code**:

**Error**:

![image](https://user-images.githubusercontent.com/79344614/108579481-48f70880-7327-11eb-8fb3-05cbaa231ae0.png)
![image](https://user-images.githubusercontent.com/79344614/108579495-5ca26f00-7327-11eb-921c-82996f6ec497.png)

### Environment
**OS System**: 
**manim version**: master 
**python version**:

abu-outis:
Have you solved it?I recently came across the same issue in my application.

PR #1543, opened 2021-06-15
Fixed the default size of Text

### LINKED ISSUE #1497: Manim write text out of the window
### Describe the error
When I use the Scene TextExample found [here](https://3b1b.github.io/manim/getting_started/example_scenes.html), only part of the text show up. I need to zoom out to see everything, and when I save the video, I can only see like I haven't zoom out.

### Code and Error
**Code**:
[This](https://3b1b.github.io/manim/getting_started/example_scenes.html) Scene TextExample.

**Error**:
Can't see some words.

### Environment
**OS System**: Windows 10
**manim version**: Release 1.0.0
**python version**: 3.9.4

### LINKED ISSUE #1533: Tex text not the correct size
### Describe the error
When running the code, the size doesn't match the one expected (and quite a bit)

### Code and Error
**Code**:
```
#Made by JsonV on the Manim Discord
class NumGenerator(Scene):
    def construct(self):
        num_gen_tex = Tex("Number Generator: ").to_edge(UL)
        self.play(Write(num_gen_tex))

        i = 0
        nums = []
        animations = []
        for k in range(10):
            a = np.random.normal(loc=100, scale=15)
            num = DecimalNumber(num_decimal_places=0).set_value(a)
            num.next_to(num_gen_tex, RIGHT)
            nums.append(num)
            animations.append(Write(num))
            animations.append(num.animate.to_edge(DL).shift(RIGHT*i))
            i += 1.3

        for animation in animations:
            self.play(animation)
        for num in nums:
            num.set_fill(ORANGE)
        nums[0].set_fill(BLUE)

        self.wait()
```

### Output
**Result Expected**
https://user-images.githubusercontent.com/73761711/121065729-62826a00-c7c9-11eb-92c2-ee10d007166f.mp4

**Result Get**
https://user-images.githubusercontent.com/73761711/121065787-6f06c280-c7c9-11eb-8ae9-69710c64889d.mp4

### Environment
**OS System**: Windows 10
**manim version**: master 
**python version**: 3.9.5

TonyCrane:
## Motivation

In my opinion, `Text` should be the same size as `TexText`.
But after the introduction of `manimpango` in #1383 , it seems that the size of `Text` is different on different machines.
For this problem, there are some issues: [ManimCommunity/ManimPango#52](https://github.com/ManimCommunity/ManimPango/issues/52), #1399, #1405, #1497, #1533.

My idea is to compare the height of a `Text` letter and `TexText` letter when rendering a scene with `Text` for the first time to get the `scale_factor` required by `Text`.
I think this idea may not be the best, but it does ensure that the size of `Text` is consistent with `TexText` on different machines. And I have come up with no other good solutions.
In addition, I have doubts about the reason why the system DPI affects the size of the text. I think this may not be a problem within manimgl?

Closes #1497.
Closes #1533.

## Proposed changes

- `manimlib/mobject/svg/text_mobject.py`: fixed the size of Text on different machines
- `docs/`: docs has been changed for this change

## Test

**Code**:
```python
class Test1(Scene):
    def construct(self):
        tex = TexText("TexText")
        text = Text("Text")
        VGroup(tex, text).arrange(RIGHT, aligned_edge=DOWN)
        self.add(tex, text)
```

**Result**:
image


manimlib/mobject/svg/text_mobject.py
```diff
@@ -66,8 +66,21 @@ def __init__(self, text, **config):
...
+            self.load_scale_factor()
+            self.scale(self.scale_factor * self.font_size)
+    
+    def load_scale_factor(self):
+        factor_file = os.path.join(get_temp_dir(), "text_scale_factor.txt")
+        if os.path.exists(factor_file):
+            with open(factor_file, 'r') as f:
+                self.scale_factor = float(f.read())
+        else:
+            tex_height = TexText("b", font_size=1).get_height()
```
naveen521kk, line 79: I worry this would mean that for rendering text LaTeX would be a must. Also, there should a better solution than this. Both the size needn't need to match but they should be made to fit into the screen.

TonyCrane: Since the size of `TexText` is the same on different machines, `tex_height` can be changed to a single number.
And if both the size matched, they are ensured to fit into the screen.

3b1b: I'm inclined to agree there's probably a better solution. Why is the right answer not to simply adjust what the constant is? Will the scale factor be different on different machines?

naveen521kk: > Will the scale factor be different on different machines?

No, I don't think so. I think adjusting the scale factor should do the trick.

Another possible solution I could think of is letting Pango handle font size in the SVG file(the `size` parameter there) and improving the SVG parser to consider the size or view box of the SVG file and not worry about scaling it. This would require a lot of work though.

TonyCrane: But in [ManimCommunity/ManimPango#52](https://github.com/ManimCommunity/ManimPango/issues/52), @GZTimeWalker proved that.
And on my old computer (Windows10, 1080p), the default size of `Text` is the same as `TexText`. On @GZTimeWalker 's computer, the default size of `Text` is four times larger than `TexText`. On my new computer (macOS 10.15, 3072x1920), the default size of `Text` is about two times larger than `TexText`.
So I think the scale factor will be different on different machines.

TonyCrane: > Another possible solution I could think of is letting Pango handle font size in the SVG file(the size parameter there) and improving the SVG parser to consider the size or view box of the SVG file and not worry about scaling it. This would require a lot of work though.

I think that's a good solution, that is to solve it in manimpango, which is also what @GZTimeWalker tends to.

naveen521kk: > And on my old computer (Windows10, 1080p), the default size of Text is the same as TexText. On @GZTimeWalker 's computer, the default size of Text is four times larger than TexText. On my new computer (macOS 10.15, 3072x1920), the default size of Text is about two times larger than TexText.

I don't think it is correct to compare `TexText` vs `Text`. Both can have different size but it needs to be scaled correctly when it is added to Manim. 
Also, I don't think Pango renders to a different size on different computers, I have tried then on Ubuntu and Windows, and get a very similar SVG file, so I think this is something to do with Manim's SVG parsing.
[svg.zip](https://github.com/3b1b/manim/files/6660556/svg.zip)

> ...that is to solve it in manimpango

That solution is to be implemented in `SVGMobject` and not in ManimPango. ManimPango already has the implementation for adjusting font size.

naveen521kk: I have something locally which may work. As in I didn't touch `SVGMobject`. Will make a PR soon so that you can test.

naveen521kk: @TonyCrane can you test https://github.com/3b1b/manim/pull/1553 to whether that works for you? At least, the text shouldn't go out of the frame.

3b1b:
After thinking about it a bit more, I think this makes sense to merge as a temporary fix. I agree it's not perfect, and in some ways its uncomfortably hacky, but it is desirable to have Tex match Text, and until someone implements a better solution we might as well have it here.

manimlib/mobject/svg/text_mobject.py
TonyCrane: Sorry to see it so late. I just tested it and the results are very good. I think that is a good solution, so I plan to revert this pr, which will cause more problems.

Issue #1385, opened 2021-02-12
Bug in Interactive Mobjects


sahilmakhijani:
### Bug
When running `ControlsExample` Scene in ExampleScenes.py, I am getting maximum recursion depth exceeded error.

**Code**:
`ControlsExample` Scene in `example_scenes.py`.

**Wrong display or Error traceback**:
```RecursionError: maximum recursion depth exceeded```

### Analysis
This bug is caused because of the following:
1. `text.add_updater(text_updater)` calls `text_updater` (because add_updater has call_updater default to True)
2. `text_updater` calls `old_text.become(new_text)`
3. `become` calls `align_family`
4. `align_family` calls `add_n_more_submobjects`
5. `add_n_more_submobjects` calls `self.copy()`
6. `copy()` calls `match_updaters`
7. `match_updaters` calls `add updater(text_updater)` on `copy_mobject`
8. `add_updater(text_updater)` calls `text_updater` (because add_updater has call_updater default to True)
9. This takes back to 1. and this continues indefinately

I tried to figure out which commit broke this Scene, and found that before this commit https://github.com/3b1b/manim/commit/d7f3f5ad24b22ed7bc85ee27df03319498fa863f, everything was working fine.
So this bug is caused due to something changed in https://github.com/3b1b/manim/commit/d7f3f5ad24b22ed7bc85ee27df03319498fa863f commit.

**Comment by sahilmakhijani** [comment]
By adding `self.suspend_updating(recurse)` back in `clear_updaters` method of `Mobject` fixes this issue. I don't know why?
This line was removed in https://github.com/3b1b/manim/commit/d7f3f5ad24b22ed7bc85ee27df03319498fa863f commit.

But now the `Scene`'s mouse_point is off of mouse_cursor which breaks interactive mobjects that respond to mouse_cursor.
To check where the mouse_point is, try running the below scene:

```
class ShowMouseCursorLocation(Scene):
    def construct(self):
        mouse_dot = Dot()
        mouse_dot.add_updater(lambda mob: mob.move_to(self.mouse_point))
        self.add(mouse_dot)
        self.interact()
```

**Comment by sahilmakhijani** [comment]
Instead of adding `self.suspend_updating(recurse)`, you could move `self.refresh_has_updater_status()` in `clear_updaters()` below `submob.clear_updaters()` just above `return_self`.

```
def clear_updaters(self, recurse=True):
        self.time_based_updaters = []
        self.non_time_updaters = []
        if recurse:
[...]

Issue #1378, opened 2021-02-10
The `turn_animation_into_updater` function does not work properly in the master branch


widcardw:
I would like to apply a function to a square when playing the `ShowCreation` animation. However, the two animations are played one by one rather than at the same time. Here is my code. 
```python
class TestAnim(Scene):
    def construct(self):
        squ = Square().shift(LEFT*3)
        anim = ShowCreation(squ, run_time=3)
        squ.add_updater(lambda m, dt: m.shift(dt*2*RIGHT))
        self.add(squ)
        turn_animation_into_updater(anim)
        self.wait(6)
```

It runs like this in cairo-backend.
![anim1](https://user-images.githubusercontent.com/55699713/107529668-1138e400-6bf6-11eb-9c1a-2fe26584a364.gif)(It may take some time to load this gif)

Maybe something goes wrong with the function `turn_animation_to_updater`. 
![anim](https://user-images.githubusercontent.com/55699713/107529377-c5863a80-6bf5-11eb-8e36-a5ed296d637b.gif)

Also, when it comes to TexText, this problem seems somewhat strange. 
```python
class TextExample
(Scene):
    def construct(self):
        t1 = TexText("Updater").scale(2).shift(UP/2).to_edge(LEFT)
        def anim_to_center(obj, dt):
            obj.shift(np.array([
                -obj.get_center()[0] * dt * 2,0,0
            ]))
        t1.add_updater(anim_to_center)
        anim1 = Write(t1)
        self.add(t1)
        turn_animation_into_updater(anim1)
        self.wait(3)
```
Result of cairo-backend: 
![out2](https://user-images.githubusercontent.com/55699713/107531571-f2d3e800-6bf7-11eb-997b-db562d9bda1f.gif)

Result of master branch(shaders): 
![textshaders](https://user-images.githubusercontent.com/55699713/107531732-15fe9780-6bf8-11eb-8823-b903348588a5.gif)

Please take a look at this issue, and I am considering create a new method to make it work properly. 
(~~Oh My Poor English~~)

3b1b:
It should be fixed now, thanks for pointing it out.

PR #1551, opened 2021-06-17
Fix ControlsExample Scene

sahilmakhijani:
## Motivation

Fixes https://github.com/3b1b/manim/issues/1385 issue (ControlsExample Scene)

## Proposed changes

1. Add `self.suspend_updating(recurse)` line back to `Mobject.clear_updaters` function
(This was removed back in https://github.com/3b1b/manim/commit/d7f3f5ad24b22ed7bc85ee27df03319498fa863f commit which fixed https://github.com/3b1b/manim/issues/1378 issue)
2. Unrelated to ControlsExample Scene, Remove unused `listen_to_events` `CONFIG` of `Mobject` which I added during Interactive Mobjects development but never used it! (https://github.com/3b1b/manim/pull/1323, https://github.com/3b1b/manim/pull/1326, https://github.com/3b1b/manim/pull/1335)

## Comments
I am not sure of the consequences of the first change. Since it was intentionally removed before to fix the https://github.com/3b1b/manim/issues/1378 issue, please check whether it is the right way of solving this issue.

### DIFF
--- a/example_scenes.py
+++ b/example_scenes.py
@@ -660,7 +660,9 @@ class ControlsExample(Scene):
                 )
             else:
                 new_text.set_opacity(0)
+            old_text.suspend_updating(recurse=False)
             old_text.become(new_text)
+            old_text.resume_updating(recurse=False, call_updater=False)
 
         text.add_updater(text_updater)
 
--- a/example_scenes.py
+++ b/example_scenes.py
@@ -646,7 +646,7 @@ class ControlsExample(Scene):
         self.add(self.panel)
 
     def construct(self):
-        text = Text("", size=2)
+        text = Text("text", size=2)
 
         def text_updater(old_text):
             assert(isinstance(old_text, Text))
--- a/example_scenes.py
+++ b/example_scenes.py
@@ -660,9 +660,7 @@ class ControlsExample(Scene):
                 )
             else:
                 new_text.set_opacity(0)
-            old_text.suspend_updating(recurse=False)
             old_text.become(new_text)
-            old_text.resume_updating(recurse=False, call_updater=False)
 
         text.add_updater(text_updater)
 
--- a/manimlib/mobject/mobject.py
+++ b/manimlib/mobject/mobject.py
@@ -56,9 +56,7 @@ class Mobject(object):
         # Must match in attributes of vert shader
         "shader_dtype": [
             ('point', np.float32, (3,)),
-        ],
-        # Event listener
-        "listen_to_events": False
+        ]
     }
 
     def __init__(self, **kwargs):
--- a/manimlib/mobject/mobject.py
+++ b/manimlib/mobject/mobject.py
@@ -514,7 +514,6 @@ class Mobject(object):
         return self
 
     def clear_updaters(self, recurse=True):
-        self.suspend_updating(recurse)
         self.time_based_updaters = []
         self.non_time_updaters = []
         self.refresh_has_updater_status()
--- a/manimlib/mobject/mobject.py
+++ b/manimlib/mobject/mobject.py
@@ -516,6 +516,7 @@ class Mobject(object):
         return self
 
     def clear_updaters(self, recurse=True):
+        self.suspend_updating(recurse)
         self.time_based_updaters = []
         self.non_time_updaters = []
         self.refresh_has_updater_status()


sahilmakhijani:
> ```python
> class TextExample(Scene):
>     def construct(self):
>         t1 = TexText("Updater").scale(2).shift(UP/2).to_edge(LEFT)
>         def anim_to_center(obj, dt):
>             obj.shift(np.array([
>                 -obj.get_center()[0] * dt * 2,0,0
>             ]))
>         t1.add_updater(anim_to_center)
>         anim1 = Write(t1)
>         self.add(t1)
>         turn_animation_into_updater(anim1)
>         self.wait(3)
> ```
https://github.com/3b1b/manim/issues/1378

- Current 3b1b/manim master branch

https://user-images.githubusercontent.com/49190295/122342844-afe99e80-cf62-11eb-8e08-0cc10729bfd5.mp4

- After adding `self.suspend_updating(recurse)` line back to `Mobject.clear_updaters` function

https://user-images.githubusercontent.com/49190295/122342779-9ea09200-cf62-11eb-9dfa-4fd86788b820.mp4


The first change of this pr breaks the above example of https://github.com/3b1b/manim/issues/1378 issue.

sahilmakhijani:
> > ```python
> > class TextExample(Scene):
> >     def construct(self):
> >         t1 = TexText("Updater").scale(2).shift(UP/2).to_edge(LEFT)
> >         def anim_to_center(obj, dt):
> >             obj.shift(np.array([
> >                 -obj.get_center()[0] * dt * 2,0,0
> >             ]))
> >         t1.add_updater(anim_to_center)
> >         anim1 = Write(t1)
> >         self.add(t1)
> >         turn_animation_into_updater(anim1)
> >         self.wait(3)
> > ```
> 
> #1378
> 
> * Current 3b1b/manim master branch
> * After adding `self.suspend_updating(recurse)` line back to `Mobject.clear_updaters` function
> The first change of this pr breaks the above example of #1378 issue.

I guess the last commit of the current master branch https://github.com/3b1b/manim/commit/5ff8e28ba573ff0365c7f685e476fac1b7113197 prefers the second one.
If so, this pr should be good to go.

TonyCrane:
Sorry to see this too late. 
After this pr, `ControlsExample` works well, but #1378 come out again.
I tried https://github.com/3b1b/manim/commit/5ff8e28ba573ff0365c7f685e476fac1b7113197 , the example in #1378 still works fine. So I think there may be anothor solution which can solve both issues?
I'm looking into it.

sahilmakhijani:
Oops, I figured out why this bug is occurring!

In `Mobject.copy()` method, submobjects are being added using the following code
```python
copy_mobject.add(*[sm.copy() for sm in self.submobjects])
```

This is problematic because submobjects also contain self, causing infinite recursion of copy().
In other places, if conditions are placed properly to check if it is not self.
Adding if check here too this fixes this error.

```python
copy_mobject.add(*[sm.copy() for sm in self.submobjects if sm is not self])
```

sahilmakhijani:
There seems to be a deeper problem here; I am unable to call `Mobject.become()` inside any updater.

For now, calling `Mobject.suspending_updating` before `Mobject.become()` inside updater fixes this issue. I have to call `resume_upddating(call_updater=False)` to resume execution of the updater.

This fixes `ControlsExample(Scene)` and doesn't touch any manim internals, so https://github.com/3b1b/manim/issues/1378 works fine.

TonyCrane:
> There seems to be a deeper problem here; I am unable to call `Mobject.become()` inside any updater.

Actually, most of the time `Mobject.become` can work in updater.
Such as: 
```python
class Code1551_1(Scene):
    def construct(self):
        mob = TexText("")
        mob.add_updater(lambda mob: mob.become(TexText("text")))
        self.add(mob)

class Code1551_2(Scene):
    def construct(self):
        mob = Square()
        mob.add_updater(
            lambda mob: mob.become(
                VGroup(Square(), Square()).arrange(RIGHT)
            )
        )
        self.add(mob)
```
and even:
```python
class Code1551_3(Scene):
    def construct(self):
        mob = Text("abc")
        mob.add_updater(lambda mob: mob.become(Text("de")))
        self.add(mob)
```
**But** if the updater make an empty `Text` to `become` an unempty `Text`, it will cause what happened in #1385 :
```python
class Code1551_4(Scene):
    def construct(self):
        mob = Text("")
        mob.add_updater(lambda mob: mob.become(Text("de")))
        self.add(mob)
```
So I think this bug may be owed to the structure of `Text`, not `ControlsExample` or interactive mobject.
I modified the `ControlsExample` to the following, and it also works fine :
```python
class ControlsExample(Scene):
    def setup(self):
        ...
        # remain the same

    def construct(self):
        text = Text("text", size=2).   # <=== here

        def text_updater(old_text):
            assert(isinstance(old_text, Text))
            new_text = Text(self.textbox.get_value(), size=old_text.size)
            # new_text.align_data_and_family(old_text)
            new_text.move_to(old_text)
            if self.checkbox.get_value():
                new_text.set_fill(
                    color=self.color_picker.get_picked_color(),
                    opacity=self.color_picker.get_picked_opacity()
                )
            else:
                new_text.set_opacity(0)
            old_text.become(new_text)

        text.add_updater(text_updater)

        self.add(MotionMobject(text))

        self.textbox.set_value("Manim")
```

TonyCrane:
And I think this bug in `Text` may not need to be fixed?
Few people will make an empty `Text` become a non-empty `Text` in the updater. Even so, they can avoid this bug by setting the initial `Text` to a non-empty `Text`.
Maybe there will be time in the future to solve this problem completely, but it is not urgent at the moment.
So I think that changing `text` to non-empty `Text` (such as `Text("text", size=2)` as I wrote) in this pr will be a simple solution to #1385.

TonyCrane (approved):

sahilmakhijani:
Aha!, I see now what's the problem. I had literally spent way too much time debugging this along with understanding manim internals and how manim work!!. Thanks a lot. I've updated the pr to fit your solution. This is the best solution which is possible right now.

Issue #1331, opened 2021-02-01
Incorrect triangulation of polygons with many holes


TonyCrane:
I tried to use `Text` to write Chinese, but there were many problems. 
For example, the simplest one is to write the word "回". But the result is this (contains unnecessary connections):
```python
class TextBug(Scene):
    def construct(self):
        txt = Text("回", font="庞门正道标题体", height=FRAME_HEIGHT-1)
        self.add(txt)
```
![result](https://user-images.githubusercontent.com/44120331/106481642-98e06d80-64e7-11eb-8ed7-fe96119374f2.png)
I lowered its opacity and showed the point set and its indices. The result is this: 
```python
class TextBug(Scene):
    def construct(self):
        txt = Text("回", font="庞门正道标题体", height=FRAME_HEIGHT-1)
        txt.set_fill(opacity=0.3)
        self.add(txt)
        points = txt[0].get_points()
        ind = index_labels(points).set_color(PURPLE).set_stroke(width=1)
        self.add(ind)
```
![TextBug](https://user-images.githubusercontent.com/44120331/106481680-a269d580-64e7-11eb-8b5d-e56e37d5cae9.png)
The position and order of the point set are correct. So I suspect that there may be an error in the triangulation, so I displayed the inner triangles in the triangulation, and the result is this: 
```python
class TextBug(Scene):
    def construct(self):
        txt = Text("回", font="庞门正道标题体", height=FRAME_HEIGHT-1)
        txt.set_fill(opacity=0.3)
        self.add(txt)
        points = txt[0].get_points()
        ind = index_labels(points).set_color(PURPLE).set_stroke(width=1)
        self.add(ind)
        tri = txt[0].get_triangulation()
        tri_indices = tri[len(points):]
        a = tri_indices[0::3]
        b = tri_indices[1::3]
        c = tri_indices[2::3]
        for i, _ in enumerate(a):
            self.add(Polygon(points[a[i]], points[b[i]], points[c[i]],\
            stroke_color=WHITE, stroke_width=4, fill_color=ORANGE, fill_opacity=0.5))
```
![TextTri](https://user-images.githubusercontent.com/44120331/106482424-68e59a00-64e8-11eb-80b9-c69e4d0d37e3.png)
Basically it is certain that there is a problem, but I don't know how to solve it. 

---
In addition, this is the svg file of this Chinese character, you can use `SVGMobject` to test it.
[...]

PR #1552, opened 2021-06-18
Temporary hack for showing text correctly

Wallbreaker5th:
## Motivation
The function `earclip_triangulation` works incorrectly with polygons with many holes. It is mentioned in #1331, but not fixed completely. The function goes wrong mainly for two reasons:
- Simply choose the closest pair of points may fail with characters that contain a ◫-like structure in it. In the following picture, the two blue edges will be added first, and the dot (outside the large square) will connect with either of the small squares (inside the large square). This happens with 疆 in the SimHei font.
  Image
- Because the first and the last point of a ring is the same, the connected ring might be self-crossed at this point (as the following gif shows). It's usually caused by a 田-like structure. This happens with 画 in the SimHei font.
  Explain2
  The red squares are two rings passed to the function (and there are more rings). The white stroke means part of the connected ring. The animation shows the order of the points on the stroke.

## Proposed changes
- It seems to be hard to write a correct `earclip_triangulation` function. I come up with several ideas, but they all have high time complexity. So I just used `insert_n_curves` to avoid the situation in the picture above. It works well most of the time, but still gets few characters wrong such as 🧬(U+1F9EC). It doubles the time taken for rendering the picture below.
- Just moving the first and the last points in the ring slightly solves the second problem.

## Test
**Code**:
```python
from manimlib import *
import time

class FontTest(Scene):
    def construct(self):
        start=time.time()

        chinese_text="疆域 画面 藤蔓 寒颤 回复 困难 \n\n蛐蛐 量取 魓 圖 轟 畾 體 團 垇 儽"
        text1 = Text(chinese_text, font="SimSun").scale(0.7).shift(UP*3)
        text2 = Text(chinese_text, font="SimHei").scale(0.7).next_to(text1,DOWN)
        text3 = Text(chinese_text, font="YouYuan").scale(0.7).next_to(text2,DOWN)
        emoji = Text("😀🛒🤯👾🙈🤞🦔🦚🐞💐🚧🧮🏁\n\n🍇🈚🥦🍔🍟🍱🎂🏺🏥🎡🥅🧬📱").next_to(text4,DOWN)
        self.add(text1, text2, text3, text4, emoji)

        end=time.time()
        print(end-start)
```

**Result**:
Before:
image

After:
image

### DIFF
--- a/manimlib/mobject/svg/text_mobject.py
+++ b/manimlib/mobject/svg/text_mobject.py
@@ -68,6 +68,9 @@ class Text(SVGMobject):
         # anti-aliasing
         if self.height is None:
             self.scale(TEXT_MOB_SCALE_FACTOR * self.font_size)
+        
+        for i in self.submobjects:
+            i.insert_n_curves(len(i.get_points()))
 
     def remove_empty_path(self, file_name):
         with open(file_name, 'r') as fpr:
--- a/manimlib/mobject/svg/text_mobject.py
+++ b/manimlib/mobject/svg/text_mobject.py
@@ -69,6 +69,8 @@ class Text(SVGMobject):
         if self.height is None:
             self.scale(TEXT_MOB_SCALE_FACTOR * self.font_size)
         
+        # Just a temporary hack to get better triangulation
+        # See pr #1552 for details
         for i in self.submobjects:
             i.insert_n_curves(len(i.get_points()))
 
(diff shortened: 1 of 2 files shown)


3b1b:
This seems like a perfectly fine temporary fix.  Maybe just add a brief comment above the relevant lines to describe what is being done, and that it is a temporary hack.

Wallbreaker5th:
> This seems like a perfectly fine temporary fix. Maybe just add a brief comment above the relevant lines to describe what is being done, and that it is a temporary hack.

I just added some. And It seems that the character 🧬(U+1F9EC) can be rendered correctly after inserting even more curves. I'm not sure whether it's necessary to turn the number of curves into part of the CONFIG.

TonyCrane (approved):
I noticed that there is a conflict. I think we should merge #1555 first and then solve the conflict in this pr before we merge this.

PR #1553, opened 2021-06-18
Scale Text Correctly

naveen521kk:
## Motivation
`Text` wasn't scaled correctly previous with https://github.com/3b1b/manim/pull/1383

Closes #1497.
Closes #1533.

## Proposed changes
Change `TEXT_MOB_SCALE_FACTOR` value
Also deprecate `size` parameter

Also, now text isn't scaled when increasing font size
instead it is passed to the underlying enging
for handling. Though always a Text object is scaled
with a default value so that it fits the screen.

## Test

Locally I get the following with this scene. It would nice if others test this.
**Code**:
```py
from manimlib import *

ipsum = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."
class TextExample(Scene):
    def fade_out_in(self, m , m1):
        self.play(FadeOut(m),FadeIn(m1),run_time=3)
    def construct(self):
        t = Text("Hello World")
        t1 = Text("Hello World \n new line")
        t2 = Text("Hello Cursive", font="Cursive")
        t3 = Text("Hello Cursive", font="Cursive", t2c = {'Hello':RED})
        t4 = Text("Hello Cursive", font="Cursive", t2c = {'Hello':RED,'Cursive':BLUE})
        t5 = Text("Hello Cursive", font="Cursive", t2c = {'Hello':RED,'Cursive':BLUE}, font_size=60)        
        t6 = Text(ipsum, font_size=30).scale(.7)
        self.add(t)
        self.fade_out_in(t,t1)
        self.fade_out_in(t1,t2)
        self.fade_out_in(t2,t3)
        self.fade_out_in(t3,t4)
        self.fade_out_in(t4,t5)
        self.fade_out_in(t5,t6)
```

**Result**:

https://user-images.githubusercontent.com/49693820/122561514-95e1b600-d05f-11eb-8367-54530dbe8abc.mp4

PR #1554, opened 2021-06-18
Remove scaline by self.font_size so that it doesn't happen twice

3b1b:
@TonyCrane, after running your fix to the text height, it looked like text was getting scaled by the font size twice.  This PR fixes things for me, but can you verify that this branch works appropriately for you? It's possible I may misunderstand something about when font_size is used for Text.

### DIFF
--- a/manimlib/mobject/svg/text_mobject.py
+++ b/manimlib/mobject/svg/text_mobject.py
@@ -19,7 +19,8 @@ from manimlib.utils.directories import get_downloads_dir, get_text_dir, get_temp
 from manimpango import PangoUtils
 from manimpango import TextSetting
 
-DEFAULT_LINE_SPACING_SCALE = 0.3
+DEFAULT_LINE_SPACING_SCALE = 0.4
+
 
 class Text(SVGMobject):
     CONFIG = {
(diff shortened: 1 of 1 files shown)


TonyCrane:
I have just tested #1553 , I think that fix is enough. It works perfectly, and there is no need for #1543 .
Now we merged both #1553 and #1543, so it works not good and need this pr to fix it.
But if we only merge #1553, it will work good.

---
Here is the code I tested for `Text`:
```python
from manimlib import *

class Test1(Scene):
    def construct(self):
        tex = TexText("TexText")
        text = Text("Text")
        VGroup(tex, text).arrange(RIGHT, aligned_edge=DOWN)
        self.add(tex, text)

class Test2(Scene):
    def construct(self):
        tex = TexText("TexText", font_size=20)
        text = Text("Text", font_size=20)
        VGroup(tex, text).arrange(RIGHT, aligned_edge=DOWN)
        self.add(tex, text)

class Test3(Scene):
    def construct(self):
        text = Text("""
            line1line1line1
            line2line2line2
            line3line3line3
        """)
        text2 = Text("line4").next_to(text, DOWN)
        self.add(text, text2)

class Test4(Scene):
    def construct(self):
        text = Text("a"*100)
        self.add(text)
```

PR #1557, opened 2021-06-19
Add MarkupText

naveen521kk:
## Motivation
Add MarkupText which is previously in Community Edition
This would use a Pango specific markup that looks like HTML.

## Proposed changes
There are some specific implementation here about `<color>`
and `<gradient>`
Pango doesn't support `<gradient>`  or `<color> ` instead it works with `color` attribute and the gradient isn't supported.
Since `SVGMobject` doesn't know about parsing colours from SVG image and implementation of `<color>` and `<gradient>` is added.

- Add a new class MarkupText

## Comments
Most parts of the code were authored by PhilippImhof and copied from Manim Community. I have added him as co-author in that commit

## Test

**Code**:
```py
class TextExample(Scene):
    def construct(self):
        a1 = MarkupText('<span style="italic">italic</span>')
        a2 = MarkupText('<span weight="bold">bold</span>')
        a3 = MarkupText('<span font="Roboto">Roboto Font</span>')
        a4 = MarkupText('<span font="Cursive">Cursive Font</span>')
        a5 = MarkupText('<color col="RED"><span font="Cursive">Red</span></color>')
        a6 = MarkupText('<gradient from="RED" to="BLUE">Gradient</gradient>')
        v = VGroup(a1, a2, a3, a4, a5, a6).arrange_in_grid()
        self.play(FadeIn(v))
```

**Result**:

https://user-images.githubusercontent.com/49693820/122654124-c6604780-d166-11eb-8374-9921adb86ead.mp4

### DIFF
--- a/manimlib/mobject/svg/text_mobject.py
+++ b/manimlib/mobject/svg/text_mobject.py
@@ -2,8 +2,12 @@ import copy
 import hashlib
 import os
 import re
+import io
 import typing
 import warnings
+import xml.etree.ElementTree as ET
+import functools
+
 from contextlib import contextmanager
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


TonyCrane (approved):

naveen521kk:
@3b1b does this look good to you?

Issue #1403, opened 2021-02-21
pip install -e . cd'd inside manim folder error


Thattenderness:
Running the pip install -e .  when I am cd'd in the unzipped manim folder I downloaded I get this error
![image_2021-02-20_225200](https://user-images.githubusercontent.com/79391411/108615912-3e437e80-73ce-11eb-9887-a12e772f3cb8.png)

C:\Manim>pip install -e .
Obtaining file:///C:/Manim
    ERROR: Command errored out with exit status 1:
     command: 'C:\Users\thatt\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\python.exe' -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Manim\\setup.py'"'"'; __file__='"'"'C:\\Manim\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base 'C:\Users\thatt\AppData\Local\Temp\pip-pip-egg-info-qf0wjr4b'
         cwd: C:\Manim\
    Complete output (16 lines):
    ERROR:root:Error parsing
    Traceback (most recent call last):
      File "c:\manim\.eggs\pbr-5.5.1-py3.9.egg\pbr\core.py", line 96, in pbr
        attrs = util.cfg_to_args(path, dist.script_args)
      File "c:\manim\.eggs\pbr-5.5.1-py3.9.egg\pbr\util.py", line 271, in cfg_to_args
        pbr.hooks.setup_hook(config)
      File "c:\manim\.eggs\pbr-5.5.1-py3.9.egg\pbr\hooks\__init__.py", line 25, in setup_hook
        metadata_config.run()
      File "c:\manim\.eggs\pbr-5.5.1-py3.9.egg\pbr\hooks\base.py", line 27, in run
        self.hook()
      File "c:\manim\.eggs\pbr-5.5.1-py3.9.egg\pbr\hooks\metadata.py", line 25, in hook
        self.config['version'] = packaging.get_version(
      File "c:\manim\.eggs\pbr-5.5.1-py3.9.egg\pbr\packaging.py", line 874, in get_version
        raise Exception("Versioning for this project requires either an sdist"
[...]

Issue #1390, opened 2021-02-14
pip install -e . doesn't work + 'manimgl' is not recognized as an internal or external command,


abartoha:
I tried it last night and it worked perfectly, but after a restart it doesn't work, says:

`manimgl is not a recocgnized internal or external command`

SO I delete everything, reclone it and do:

### Code and Error
`pip install -e .`

and I get:
```
WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue.
To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.
Defaulting to user installation because normal site-packages is not writeable
Obtaining file:///C:/Users/USER/Desktop/animation/manim
    ERROR: Command errored out with exit status 1:
     command: 'c:\program files\python38\python.exe' -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Users\\USER\\Desktop\\animation\\manim\\setup.py'"'"'; __file__='"'"'C:\\Users\\USER\\Desktop\\animation\\manim\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base 'C:\Users\USER\AppData\Local\Temp\pip-pip-egg-info-gbw01nwl'
         cwd: C:\Users\USER\Desktop\animation\manim\
    Complete output (16 lines):
    ERROR:root:Error parsing
    Traceback (most recent call last):
      File "c:\users\user\desktop\animation\manim\.eggs\pbr-5.5.1-py3.8.egg\pbr\core.py", line 96, in pbr
        attrs = util.cfg_to_args(path, dist.script_args)
      File "c:\users\user\desktop\animation\manim\.eggs\pbr-5.5.1-py3.8.egg\pbr\util.py", line 271, in cfg_to_args
        pbr.hooks.setup_hook(config)
      File "c:\users\user\desktop\animation\manim\.eggs\pbr-5.5.1-py3.8.egg\pbr\hooks\__init__.py", line 25, in setup_hook
        metadata_config.run()
      File "c:\users\user\desktop\animation\manim\.eggs\pbr-5.5.1-py3.8.egg\pbr\hooks\base.py", line 27, in run
        self.hook()
[...]

PR #1565, opened 2021-06-26
Transition build method to PEP 517

TonyCrane:
## Motivation

Because manim always use `pbr` for setup, which needs to automatically obtain the version from the `.git/` folder. This leads to the issues of #1390 #1403 that cannot be installed from the `.zip` file.
So I transitioned to the method of using `pyproject.toml` and `setup.cfg` to build, which is also recommended by `setuptools`.

## Proposed changes

- D `setup.py`
- M `setup.cfg`
- N `pyproject.toml`
- N `MANIFEST.in`

## Test

https://test.pypi.org/project/manimgl/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,0 @@
-#!/usr/bin/env python
-
-from setuptools import setup
-
-setup(
-    setup_requires=['pbr'],
-    pbr=True,
-)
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -1,8 +0,0 @@
-#!/usr/bin/env python
-
-from setuptools import setup
-
-setup(
-    setup_requires=['pbr'],
-    pbr=True,
```
naveen521kk, line 7: I think having a setup.py will not hurt. Just remove the last argument about pbr.

PR #1566, opened 2021-06-29
Add frame to the scene when initialization

pdcxs:
If we add an updater to the frame of the camera, and have not added the frame into the scene before, the updater will not work. So, I suggest to add the frame to the objects of the scene at the initilization stage.


TonyCrane:
But when we call `Scene.clear()`, `self.mobjects` will still become an empty list.

pdcxs:
> But when we call `Scene.clear()`, `self.mobjects` will still become an empty list.

I think it's ok, when we fork clear method, I think it's reasonable to clear the frame updater at the same time

TonyCrane (approved):

PR #1578, opened 2021-07-14
Minor fix for zooming

nutanstrek:
## Motivation

It's weird that when you Scroll Up, the interactive shell zooms out, this is generally the other way round for most of the UI's.

## Proposed changes

Simply replace factor -> 1 / factor.

### DIFF
--- a/manimlib/scene/scene.py
+++ b/manimlib/scene/scene.py
@@ -570,7 +570,7 @@ class Scene(object):
         frame = self.camera.frame
         if self.window.is_key_pressed(ord("z")):
             factor = 1 + np.arctan(10 * offset[1])
-            frame.scale(factor, about_point=point)
+            frame.scale(1/factor, about_point=point)
         else:
             transform = frame.get_inverse_camera_rotation_matrix()
             shift = np.dot(np.transpose(transform), offset)


TonyCrane:
**I guess** this problem is caused by the _natural scrolling_ of the Mac system.

Grant uses MacOS, and the default scrolling in MacOS is _natural scrolling_, which makes the operation on the touchpad very natural and direct. But if MacOS is connected to an ordinary mouse, the scrolling direction of the mouse wheel will be opposite to that of Windows (Windows is using _unnatural scrolling_, and the direction is just opposite to that of Mac). At this time, a software is generally installed to reverse the scrolling direction of the mouse. In this case, scrolling up with the mouse is equivalent to scrolling down under normal conditions (including Windows and Mac). Therefore, Grant may be designed like this, so that the mouse on the Mac can achieve the effect of scrolling up to zoom in, but it has caused the problem of scrolling up to zoom out on other systems.

I don’t know if it’s a good idea to judge the OS here, maybe @3b1b should come and have a look?

> I am also using a Mac system. I installed the `Scroll Reverser` software to reverse the direction of the mouse wheel. In other places, I can use the mouse wheel as on the Windows system, and scroll up when I press `z` to zoom in the scene. So I made this conjecture.

3b1b:
Honestly, either one works for me.

PR #1591, opened 2021-08-03
Improve `Mobject.scale`

TonyCrane:
## Motivation

Previous `Mobject.scale` method doesn't use overrided method, which causes some problems when scaling a VGroup:

**Code**:
```python
class Code(Scene):
    def construct(self):
        num = VGroup(
            DecimalNumber(1), 
            Text("foo")
        ).scale(4).arrange(RIGHT)
        self.add(num)
        self.play(num[0].animate.set_value(2))
```
**Output**:

https://user-images.githubusercontent.com/44120331/127976196-03ea964f-3cf5-4260-afa0-6da174f4ce28.mp4

Because `DecimalNumber` override `scale` method to change the `font_size` attribute.

This PR is implemented by recursively calling `scale` method on the submobjects of the current mobject.

## Proposed changes

- M `manimlib/mobject/mobject.py`: rewrote `scale` method

## Test

**Code**:
```python
class Code(Scene):
    def construct(self):
        num = VGroup(
            DecimalNumber(1), 
            Text("foo")
        ).scale(4).arrange(RIGHT)
        self.add(num)
        self.play(num[0].animate.set_value(2))
```
**Result**:

https://user-images.githubusercontent.com/44120331/127976835-cd7c9246-fb4e-4f1b-b9b2-05040d3c88b1.mp4

### DIFF
--- a/manimlib/mobject/mobject.py
+++ b/manimlib/mobject/mobject.py
@@ -561,7 +561,7 @@ class Mobject(object):
         )
         return self
 
-    def scale(self, scale_factor, min_scale_factor=1e-8, **kwargs):
+    def scale(self, scale_factor, min_scale_factor=1e-8, about_point=None, about_edge=ORIGIN, recurse=True):
         """
         Default behavior is to scale about the center of the mobject.
         The argument about_edge can be a vector, indicating which side of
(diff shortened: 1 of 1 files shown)


3b1b:
It turns out this breaks a number of other things, so I will revert it.

For this particular example, it feels like the problem has more to do with DecimalNumber tracking its font size changes in an unreliable way. I will see if there's a better way to address that.

PR #1689, opened 2021-12-02
Visual directory structure

AlexEidt:
## Motivation

I've added a visual directory structure to `docs/source` in a `README.md` file. A visual layout of the directory may help some developers better navigate this library. I also added a directory structure visualization in `docs/source/getting_started/structure.rst`.

## Proposed changes

- Added SVG images of directory structure in a graphical form.

## Test

**Code**: N/A. This is just an image in the documentation of the project.

**Result**: Note that this is rendered as PNG only for this preview. The actual are SVG.
graph


TonyCrane:
Thank you for what you did. 
But I wonder whether it makes sense to add this image, because all the meaningful files have been listed on that page of the document, and the function of each file has also been explained. 
I think that is enough.

AlexEidt:
Thank you for your consideration.

PR #1694, opened 2021-12-16
Add option to add ticks on x-axis in BarChart()

DangGiaChi:
## Motivation

It's only a minor change. I found out that BarChart() doesn't have an option to add ticks on x-axis. I know it's because the x-axis is usually for names or labels and not values, but I think it's nice to add an option to add ticks to seperate bars for better visualization.

## Proposed changes

- I looked up at how manim add ticks on y-axis and created an option for people to also add ticks to the x-axis
- I added a config parameter "include-x-ticks"
- The number of ticks divided is also the number of values

## Test

**Code**:
```py 
from manimlib import *

class Test(Scene):
    def construct(self):
        global chart
        bar_value = [50]*10
        chart = BarChart(
            bar_value,
            label_y_axis = True,
            n_ticks = 10,
            max_value = 100,
            bar_label_scale_val = 0.5,
            include_x_ticks = True
        )
        self.add(chart)
```

**Result**
Test
:

### DIFF
--- a/manimlib/mobject/probability.py
+++ b/manimlib/mobject/probability.py
@@ -8,6 +8,7 @@ from manimlib.mobject.svg.tex_mobject import TexText
 from manimlib.mobject.types.vectorized_mobject import VGroup
 from manimlib.utils.color import color_gradient
 from manimlib.utils.iterables import listify
+from numpy import ndarray
 
 EPSILON = 0.0001
 
@@ -149,7 +150,9 @@ class BarChart(VGroup):
         "height": 4,
         "width": 6,
         "n_ticks": 4,
+        "include_x_ticks": False,
         "tick_width": 0.2,
+        "tick_width_x": 0.15,
         "label_y_axis": True,
         "y_axis_label_height": 0.25,
         "max_value": 1,
@@ -165,6 +168,7 @@ class BarChart(VGroup):
         if self.max_value is None:
             self.max_value = max(values)
 
+        self.n_ticks_x = len(values)
         self.add_axes()
         self.add_bars(values)
         self.center()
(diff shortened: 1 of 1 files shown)


manimlib/mobject/probability.py
```diff
@@ -8,6 +8,7 @@
 from manimlib.mobject.types.vectorized_mobject import VGroup
 from manimlib.utils.color import color_gradient
 from manimlib.utils.iterables import listify
+from numpy import ndarray
```
TonyCrane, line 11: No need for this

manimlib/mobject/probability.py
```diff
@@ -182,6 +186,17 @@ def add_axes(self):
             ticks.add(tick)
         y_axis.add(ticks)
 
+        if self.include_x_ticks == True:
+            ticks1 = VGroup()
+            widths = np.linspace(0, self.width, self.n_ticks_x + 1)
+            label_values = np.linspace(0, len(self.bar_names), self.n_ticks_x + 1)
+            for x, value in zip(widths, label_values):
+                tick1 = Line(UP*0.05, DOWN*0.05)
+                tick1.set_width(self.tick_width_x)
```
TonyCrane, line 195: Ticks on the y axis can use `set_width` because they are horizontal.
But here, `set_width` won't make sense.
And why you named this local variable to `tick1`? `tick` is also available.
These two lines should be modified to:
```py
tick = Line(DOWN, UP)
tick.set_height(self.tick_width_x) # maybe call it `tick_height` or `tick_length_x` is better?
```

TonyCrane (changes requested):
Here are some suggestions. Hope you can improve them :)

DangGiaChi:
Okok, I'll check it out as soon as it's morning at my place:)

manimlib/mobject/probability.py
DangGiaChi: They're mainly because I was confused at what each line does at first:)

manimlib/mobject/probability.py
```diff
@@ -187,15 +186,14 @@ def add_axes(self):
...
             for x, value in zip(widths, label_values):
-                tick1 = Line(UP*0.05, DOWN*0.05)
-                tick1.set_width(self.tick_width_x)
+                tick1 = Line(UP, DOWN)
+                tick1.set_height(self.tick_height)
                 tick1.move_to(x * RIGHT)
-                ticks.add(tick1)
-            x_axis.add(ticks1)
+                ticks.add(tick)
+            x_axis.add(ticks)
```
TonyCrane, line 196: Oh, I think you need a new VGroup for `x_ticks` and add it to `x_axis`, instead of still using `ticks`.

TonyCrane (changes requested):

manimlib/mobject/probability.py
DangGiaChi: Should I also change the ticks variables names on y-axis so that the names on x and y-axis have the same format?

TonyCrane: that's better

TonyCrane (approved):

PR #1716, opened 2022-01-26
Refactor MTex and some clean-ups

YishiMichael:
## Proposed changes
- M `manimlib/mobject/svg/mtex_mobject.py`: add `generate_plain_tex_file` config option (default False) to allow users generate another svg without additional `\color` commands to prevent the shape of svg from being altered by them; some refactors for efficiency improvememt; fix the bug that multi-line tex fails rendering
- M `manimlib/utils/simple_functions.py`: refactor `choose` with lru_cache and scipy's func (borrowed from community edition); delete `clip_in_place` func (unused)
- M `manimlib/utils/color.py`: delete `get_shaded_rgb` func (unused)
- D `manimlib/utils/string.py`: clean up unused functions
- M `manimlib/mobject/coordinate_systems.py`: remove redundancies
- M `manimlib/__init__.py`: remove `manimlib/utils/string.py` import


TonyCrane (approved):

YishiMichael:
I noticed that breaking up every pair of braces may cause unexpected issues, so in the current version only tex enclosed by a pair of double-braces (`{{ ... }}`) are split out, making sure that these splitting are intended.
Note, this version only works after #1717 is merged.
MTex currently cannot handle tex originally including `\color` command. Users may use `tex_to_color_map` parameter instead.

PR #1717, opened 2022-01-26
Parse and generate style for SVG

TonyCrane:
## Motivation

Before, `SVGMobject` will remove all styles from a svg file.
This pull request parse the style for SVG by refering to [ManimCommunity/manim](https://github.com/ManimCommunity/manim).
In order to prevent the style generated layer by layer in `init_points` from being globally overwritten by the `init_colors` after it. I added a parameter `override` to `init_colors`, which is set to `True` by default (that is, consistent with the original performance), but when `override = False`, it will recurse layer by layer and set the color to the currently generated color.

## Proposed changes

- M `manimlib/mobject/svg/svg_mobject.py`: add SVG style parser and fix some bugs of path string
- M `manimlib/mobject/types/vectorized_mobject.py`: add `override` parameter to `init_colors`
- M `manimlib/mobject/mobject.py`: synchronously update the parameters of `init_colors`
- M `manimlib/mobject/geometry.py`: synchronously update the parameters of `init_colors`
- M `manimlib/mobject/types/surface.py`: synchronously update the parameters of `init_colors`
- M `manimlib/mobject/svg/tex_mobject.py`: synchronously update the parameters of `init_colors` and synchronize `SingleStringTex`'s `stroke_width` to `SVGMobject`
- M `manimlib/mobject/svg/tex_mobject.py`: cache color in `tex_string_to_mob_map` together with tex_string
- M `manimlib/mobject/svg/text_mobject.py`: replace `warnings.warn` with `log.warning`

## Test

I tested all the svg files used in ManimCE for testing and each file renders correctly with color and stroke.
And `Tex` `TexText` can correctly generate texts with color according to the corresponding LaTeX expression (`\color{}`)

**Code**:
```python
class Code(Scene):
    def construct(self):
        svg = VGroup(
            TexText(r"test {\color{red}and} {\color{blue}test}").set_width(10),
            TexText(r"test {\color{red}and} {\color{blue}test}").set_width(10).set_color(WHITE),
            TexText(r"test {\color{red}and} {\color{blue}test}", color=BLACK).set_width(10),
        ).arrange(DOWN)
        self.add(svg)
```
**Result**:
image
As can be seen from the above example, the color parameter is only the default color, and this color will be overwritten as long as a color is included in the SVG. If you want to ignore the color in SVG, you can directly `.set_color` again.

### DIFF
--- a/manimlib/mobject/geometry.py
+++ b/manimlib/mobject/geometry.py
@@ -608,8 +608,8 @@ class Arrow(Line):
         self.insert_tip_anchor()
         return self
 
-    def init_colors(self):
-        super().init_colors()
+    def init_colors(self, override=True):
+        super().init_colors(override)
         self.create_tip_with_stroke_width()
 
     def get_arc_length(self):
--- a/manimlib/mobject/mobject.py
+++ b/manimlib/mobject/mobject.py
@@ -109,8 +109,8 @@ class Mobject(object):
             "reflectiveness": self.reflectiveness,
         }
 
-    def init_colors(self):
-        self.set_color(self.color, self.opacity)
+    def init_colors(self, override=True):
+        self.set_color(self.color, self.opacity, override)
 
     def init_points(self):
         # Typically implemented in subclass, unlpess purposefully left blank
--- a/manimlib/mobject/svg/svg_mobject.py
+++ b/manimlib/mobject/svg/svg_mobject.py
@@ -4,6 +4,7 @@ import string
 import os
 import hashlib
 
+from colour import web2hex
 from xml.dom import minidom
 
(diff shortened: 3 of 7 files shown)


3b1b:
This looks great, thanks for doing it

3b1b:
There are a few examples I can find, like the one attached, where due to how the style is represented in the SVG it does not quite work, but it's certainly better than current functionality so worth the merge.

beer
.

TonyCrane:
Yes, it's because the styles are defined by the css in the <style> tag, which is not handled yet (nor is ManimCommunity). I'll try my best to handle this situation when I have time :)

PR #1744, opened 2022-02-15
Add text_config for DecimalNumber

TurkeyBilly:
Added config passing for text

## Motivation

Give users option to change the text config for ```DecimalNumber```

## Proposed changes

- Added ```text_config``` in decimal number config
- Removed extra ```init_colors()``` so that ```t2c``` can be passed in
- Adjusted```string_to_mob``` method so key word arguments can be accepted, meanwhile maintain the``` string_to_mob_map```

## Test

**Code**:
```python
class Test(Scene):
    def construct(self):
        a = DecimalNumber(1.02, show_ellipsis=True)
        b = DecimalNumber(1.02, show_ellipsis=True, 
                            text_config=dict(font="仿宋", slant=ITALIC)).shift(DOWN)
        c = DecimalNumber(1.02, show_ellipsis=True, 
                            text_config=dict(font="Times New Romen", t2c={'0':RED})).shift(2*DOWN)
        self.add(a, b, c)
```

**Result**:
image

### DIFF
--- a/manimlib/mobject/numbers.py
+++ b/manimlib/mobject/numbers.py
@@ -20,6 +20,7 @@ class DecimalNumber(VMobject):
         "include_background_rectangle": False,
         "edge_to_fix": LEFT,
         "font_size": 48,
+        "text_config": {} # Do not pass in font_size here
     }
 
     def __init__(self, number=0, **kwargs):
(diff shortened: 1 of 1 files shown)


manimlib/mobject/numbers.py
```diff
@@ -85,10 +85,15 @@ def init_data(self):
...
-        mob = string_to_mob_map[string].copy()
+    def string_to_mob(self, string, mob_class=Text, **kwargs):
+        if not kwargs:
+            if string not in string_to_mob_map:
+              string_to_mob_map[string] = mob_class(string, font_size=1, **kwargs)
+            mob = string_to_mob_map[string].copy()
+        else:
+            if (string, str(kwargs)) not in string_to_mob_map:
+              string_to_mob_map[(string, str(kwargs))] = mob_class(string, font_size=1, **kwargs)
+            mob = string_to_mob_map[(string, str(kwargs))].copy()
```
TonyCrane, line 96: Use `str` to make a dict hashable is not a good way.
And why not combine these two situations together?

TonyCrane (changes requested):

manimlib/mobject/numbers.py
TurkeyBilly: I was thinking not to combine because I am afraid to influence the speed for users who do not need to pass in  text_kwargs.
I changed the hashing way according to https://stackoverflow.com/questions/5884066/hashing-a-dictionary, how about now

TonyCrane (approved):

PR #1789, opened 2022-04-12
Sort imports

YishiMichael:
## Motivation
Sort imports in alphabetical order. Expand all imports from `from manimlib.constants import *` (except `manimlib/mobject/svg/drawings`). Refactor some methods of `LabelledString` class.

## Proposed changes
- M `manimlib/mobject/utils/color.py`: Add annotations and refactor with functions supported by `colour` package.
- M `manimlib/mobject/coordinate_systems.py`: Make `CoordinateSystem` inherit `ABC`.
- M `manimlib/mobject/svg/svg_mobject.py`: Change `hex` to `hexrgb` to prevent from passing in an 8-digit-hex to `fill_color` and `stroke_color`.
- M `manimlib/utils/iterables.py`: Support hashing `Color` objects.
- Sort imports alphabetically, remove unused imports, adjust type annotations, etc.

## Sidenote

I once decided to refactor `Mobject.scale` function through the following statement:
```py
scale_factor = np.resize(scale_factor, self.dim).clip(min=min_scale_factor)
```

This statement can simultaneously handle float and array objects, but it hurts the behavior of `DecimalNumber._handle_scale_side_effects` method, which can only work when `scale_factor` is a float. So I'd suggest some thoughts about removing `font_size` from `DecimalNumber.data` (`font_size` is even not stored in `Tex.data` or `Text.data`).


3b1b:
Thank you for doing this. For future reference, though, it would be helpful to have large changes like this with several qualitatively different types of updates separated out into different PRs. For example, the reordering of import lines touches many files, but in principle doesn't require much to review, whereas certain refactor may deserve a little more attention.

In playing around with it, I noticed at least two little errors:

- It's actually important for the line "from manimlib.window import *" to show up above the others in __init__.py, at least on my system.
- three_dimensions.py needs to import numpy as np

TonyCrane:
There is another problem with importing window if I remember correctly. 

Currently just `from manimlib import *` will cause pyglet to create the window, looking for the display device. 
Therefore, manimgl cannot currently run on machines without a display device (such as WSL and remote servers) even if the purpose is to render video rather than preview or interact.
This also resulted in manimgl not being able to write CI checks or remotely render videos.
I think maybe there should be a better way of imports to avoid this.

3b1b:
You know, there's probably no good reason for that import line to even be among the global imports. The only thing it defines is Window, which is only needed by Scene during its init, which is handled there.

YishiMichael:
> Thank you for doing this. For future reference, though, it would be helpful to have large changes like this with several qualitatively different types of updates separated out into different PRs. For example, the reordering of import lines touches many files, but in principle doesn't require much to review, whereas certain refactor may deserve a little more attention.
> 
> In playing around with it, I noticed at least two little errors:
> 
> * It's actually important for the line "from manimlib.window import *" to show up above the others in **init**.py, at least on my system.
> * three_dimensions.py needs to import numpy as np

Thanks! Already split into #1789 and #1795.

PR #1795, opened 2022-04-23
Refactor LabelledString and relevant classes

YishiMichael:
## Motivation
Refactor some methods of `LabelledString`, `Text`, `MTex` classes.

## Proposed changes
- M `manimlib/mobject/svg/labelled_string.py`: Some refactors (render plain svg file first instead of the labelled svg); remove `use_plain_tex` attributes; Support a much more flexible `Selector` type; rename some methods.
- M `manimlib/mobject/svg/mtex_mobject.py`: Allow passing in a 2-tuple to `tex_environment`, regarded as a prefix and a suffix respectively; remove `use_plain_tex` attributes.
- M `manimlib/mobject/svg/text_mobject.py`: Remove `use_plain_tex` attributes.

The method renaming in `LabelledString` class includes:
```
get_parts_by_string => select_parts
get_part_by_string => select_part
set_color_by_string => set_parts_color
set_color_by_string_to_color_map => set_parts_color_by_dict
```

The names of aliases remain unchanged (e.g. `MTex.get_parts_by_tex`)

The type definition of `Selector` is given below:
```py
Selector = Union[
    str,
    re.Pattern,
    tuple[Union[int, None], Union[int, None]],
    Iterable[Union[
        str,
        re.Pattern,
        tuple[Union[int, None], Union[int, None]]
    ]]
]
```

The `LabelledString.select_parts` method takes in a `selector` as an input. The `isolate` parameter, keys of `tex_to_color_map`, keys of `t2c`, etc. are also of type `selector`. `LabelledString.select_parts` no longer supports `case_sensitive` and `regex` parameters, but still supports `substring` parameter.

```py
# Equivalent:
text_ = Text("abc", isolate="ab")
text_ = Text("abc", isolate=["ab"])
text_ = Text("abc", isolate=(0, 2))
text_ = Text("abc", isolate=(None, -1))

# Equivalent:
text = Text("abcBC", isolate=["a", "bc", "BC"])
text = Text("abcBC", isolate=["a", re.compile("bc", re.I)])
text = Text("abcBC", isolate=[(0, 1), (1, 3), (3, 5)])
text = Text("abcBC", isolate=[(None, 1), (1, 3), (-2, None)])

# Equivalent:
g = text.select_part("bcBC")
g = text.select_part(["bc", "BC"])
g = text.select_part(re.compile("bc", re.I))
g = text.select_part((1, None))

# Equivalent:
poly = MTex(
[...]

Issue #1792, opened 2022-04-16
Cannot show text in Ubuntu20.04


qinjinbnu:
# `Text` cannot show anything.

### **Code**: 

the directory is like
```
.
├── code.py
└── custom_config.yml
```

then I ran `manimgl code.py` which shown below

```python
class OpeningManimExample(Scene):
    def construct(self):
        intro_words = Text("""
            The original motivation for manim was to
            better illustrate mathematical functions
            as transformations.
        """)
        intro_words.to_edge(UP)

        self.play(Write(intro_words))
        self.wait(2)

        # Linear transform
        grid = NumberPlane((-10, 10), (-5, 5))
        matrix = [[1, 1], [0, 1]]
        linear_transform_words = VGroup(
            Text("This is what the matrix"),
            IntegerMatrix(matrix, include_background_rectangle=True),
            Text("looks like")
        )
        linear_transform_words.arrange(RIGHT)
        linear_transform_words.to_edge(UP)
        linear_transform_words.set_stroke(BLACK, 10, background=True)

        self.play(
            ShowCreation(grid),
            FadeTransform(intro_words, linear_transform_words)
        )
        self.wait()
        self.play(grid.animate.apply_matrix(matrix), run_time=3)
        self.wait()

        # Complex map
        c_grid = ComplexPlane()
        moving_c_grid = c_grid.copy()
        moving_c_grid.prepare_for_nonlinear_transform()
        c_grid.set_stroke(BLUE_E, 1)
        c_grid.add_coordinate_labels(font_size=24)
        complex_map_words = TexText("""
            Or thinking of the plane as $\\mathds{C}$,\\\\
            this is the map $z \\rightarrow z^2$
        """)
        complex_map_words.to_corner(UR)
        complex_map_words.set_stroke(BLACK, 5, background=True)

        self.play(
            FadeOut(grid),
            Write(c_grid, run_time=3),
            FadeIn(moving_c_grid),
            FadeTransform(linear_transform_words, complex_map_words),
        )
        self.wait()
        self.play(
            moving_c_grid.animate.apply_complex_function(lambda z: z**2),
            run_time=6,
        )
        self.wait(2)
```

### **Error**:

```bash
ManimGL v1.6.1
[23:55:05] INFO     Using the default configuration file, which you can  config.py:265
                    modify in `~/anaconda3/envs/manim/lib/pyt              
                    hon3.8/site-packages/manimlib/default_config.yml`                 
[...]

Issue #1760, opened 2022-03-11
After using  math font package in tex_template of ManinGL， the TexText cannot displayed correctly，


inoxch:
### Describe the bug

After using  math font package in tex_template of ManinGL，such as  “\usepackage{fourier}”,
Once I use the latex command that involves changing  words size， the TexText will be displayed incorrectly，
such as “$x^{y}$ ”,   “ $x_{y}$”,     ”{\small $x$}+$y$".

I inspect the (.svg) file it outputs.
the result is correct.
So I think it is a displayed bug
**Code**:

**Wrong display or Error traceback**:

inoxch:
[https://github.com/inoxch/a/blob/main/11.png](url)
output
[https://github.com/inoxch/a/blob/main/2.png](url)

inoxch:
![11](https://user-images.githubusercontent.com/101418800/157966769-3a6910b9-2b22-4e32-a394-707aa146223b.png)
output：
![2](https://user-images.githubusercontent.com/101418800/157966784-199cdc64-b138-4a10-b026-599880eb2212.png)

TonyCrane:
Can you attach the .svg file ?

inoxch:
> Can you attach the .svg file ?
![1](https://user-images.githubusercontent.com/101418800/158079358-d7994da4-45b0-444e-89fa-115432b0c8e7.svg)
![2](https://user-images.githubusercontent.com/101418800/158079360-69fbab49-01e7-44b8-b900-95e88b488b3f.svg)

inoxch:
> Can you attach the .svg file ?
[svg.zip](https://github.com/3b1b/manim/files/8240644/svg.zip)

PR #1818, opened 2022-05-22
Add `template` and `additional_preamble` parameters to `Tex`

YishiMichael:
### Supports self-defined tex templates and temporary addition of preambles
Sometimes users may want to temporarily use tex commands from tex packages not included in tex template files by default. So I add support for it. This is inspired from Manim Community. For classes `Tex`, `MTex`, parameters `template` and `additional_preamble` are added. Example usages are shown below.

Demo
```python
tex = Tex(
    "a^2 + b^2 = c^2",
    template="French Cursive"
)
self.add(tex)
```
<details>
<summary>Result</summary>

TexTemplateDemo1
</details>

Demo
```python
chem = TexText(
    "\\ce{Zn^2+ <=>[+ 2OH-][+ 2H+] "
    "$\\underset{\\text{amphoteres Hydroxid}}{\\ce{Zn(OH)2 v}}$ "
    "<=>[+ 2OH-][+ 2H+]"
    "$\\underset{\\text{Hydroxozikat}}{\\ce{[Zn(OH)4]^2-}}$}",
    template="empty",
    additional_preamble="\\usepackage{mhchem}"
).scale(0.7)
self.add(chem)
```
<details>
<summary>Result</summary>

TexTemplateDemo2
</details>

Demo
```python
textext = MTexText(
    "The quick brown fox jumps over the lazy dog.",
    template="French Cursive", isolate="brown"
)
textext.select_part("brown").set_fill(LIGHT_BROWN)
self.add(textext)
```
<details>
<summary>Result</summary>

TexTemplateDemo3
</details>

58 tex templates in total are provided in `manimlib/tex_templates.yml`. Users may feel free to add whatever template they want to this file for own usage. These templates can simply be accessed by passing their names to `template` attribute. By default, `template` is an empty string, in which case manim will use the default template specified at `style.tex_template` in the configuration file. Currently this key replaces all 4 attributes under `tex` key. The original template files (`tex_template.tex` and `ctex_template.tex`) are now removed, and are now included as `default` and `ctex` for backward compatibility.

Users may add preambles specifically to a `Tex` or `MTex` instance via `additional_preamble` attribute, which is also an empty string by default.

Frankly speaking, I'm a bit worried if having so much packages imported in a tex file could cause side effects. For instance, the `"\vdots"` command doesn't show up in the final svg (not caused by questionable parsing of `svgelements`; even the path is not defined). This issue may be caused by the line `"\usepackage[T1]{fontenc}"` (which I'm not sure). The point is, I would suggest using a template with only a few elementary packages imported, and add packages for specific `Tex` instances. Manim Community is also concerned about such an issue (https://github.com/ManimCommunity/manim/issues/2889). That's why `basic`, `empty` templates show up in the `tex_templates.yml`. I also collected all 52 templates from the Community Edition and sorted them alphabetically.

### New `protect` attribute for `StringMobject`
Users may use `protect` attribute to specify some substrings that are not expected to be broken up.

Demo
```python
protected_tex = MTex(
    "\\begin{bmatrix} a & b \\\\ - b & a \\end{bmatrix}",
    tex_to_color_map={"a": GREEN, "b": RED},
    protect="{bmatrix}"
)
self.add(protected_tex)
```
<details>
<summary>Result</summary>

ProtectDemo1
</details>

In the demonstration above, without specifying `protect="{bmatrix}"`, all of letter `b`s (including ones in the word `bmatrix`) will be isolated. In fact, `MTex` can automatically handle most cases --- it protects all substrings matched by the regex expression `\\([a-zA-Z]+|.)`. The case above is tricky so that we shall tackle it manually. Sidenote, this attribute only prevents from cutting **halfway** at protected spans, so isolating `"\\begin{bmatrix}"` in the example above is still allowed (but nobody will do so, unless on purpose).

`protect` attribute should be of type `Selector`, the same as `isolate`.

### Add back `disable_ligatures` attribute in `MarkupText` and `Text`
For backward compatibility. Default to be `True`.

### Removes `is_markup` attribute for `MarkupText` and `Text`
This attribute is not expected to be altered up to the class --- so why not remove it?

### No longer allows passing in a 2-tuple to `tex_template` attribute in `MTex`
This feature was originally added in #1795 in order to escape isolating substrings in, e.g., `\begin{table}{ccc}`, `\end{table}` (which cannot be solved by passing `tex_environment="table"`). Now we can make use of `protect` attribute to overcome this problem in an elegant way.

### No longer adjusts spans in `StringMobject`
In the previous implementation, I designed an algorithm to handle spans pointing to crazy substrings, like `"a{b"`. From now on, such spans will receive warnings and will no longer be adjusted. Similarly, partly overlapping substrings now lead to warnings instead of exceptions.

### #1760 After using math font package in tex_template of ManimGL, the TexText cannot displayed correctly
This issue is caused since `svgelements` cannot parse `use` elements whose referred elements are defined afterwards. This is fixed by expanding all `use` elements in svg before parsing.

### #1792 Cannot show text in Ubuntu20.04
`Text` uses the `line_height` attribute to control the line spacing, which requires pango with version >=1.50. The version of pango is now checked to decide whether to insert this attribute.

### #1848 Fix: can't parse `<u>Boolean Operation</u>`
Fixed passingly in the refactory.

### #1850 ValueError: Invalid markup string - when entering characters in Interactive mode
The `font_size` attribute in a markup tag can only be integers. This issue is fixed by inserting a `round` function.

### Other bugs
`MTex.get_part_by_tex` and `Text.get_part_by_text` miss `**kwargs` when passing attributes.

`manimlib/animation/indication.py` file misses imports from `manimlib.constants`.

## Proposed changes
- M `manimlib/animation/indication.py`: Add missed imports.
- M `manimlib/animation/transform_matching_parts.py`: Refactor `TransformMatchingStrings`.
- M `manimlib/default_config.yml`: Add `tex_template` attribute; remove `tex` items.
- M `manimlib/mobject/svg/mtex_mobject.py`: Add `template`, `additional_preamble` attributes; refactor `MTex` according to `StringMobject`.
- M `manimlib/mobject/svg/string_mobject.py`: Refactor `StringMobject`.
- M `manimlib/mobject/svg/svg_mobject.py`: Expand `use` elements beforehand; use `hash_string` function.
- M `manimlib/mobject/svg/tex_mobject.py`: Add `template`, `additional_preamble` attributes.
- M `manimlib/mobject/svg/text_mobject.py`: Remove `is_markup` attribute; refactor `Text` according to `StringMobject`.
- A `manimlib/tex_templates.yml`: Add preambles of tex templates.
- R `manimlib/tex_templates/.`: Remove tex template files.
- M `manimlib/utils/init_config.py`: Alter according to `default_config.yml`.
- M `manimlib/utils/simple_functions.py`: Add `hash_string` function.
- M `manimlib/utils/tex_file_writing.py`: Make manim able to generate tex files based on specific preamble; remove `tex_hash` function.


TonyCrane (approved):
LGTM. But we'd better add a new example about `Tex` and `TexText` in `example_scenes.py`. And it's better to add a new page to the docs to introduce these full features of them.

YishiMichael:
Recently the issue concerned with `<use>` tags in `SVGMobject` has been handled in the [latest version](https://github.com/meerk40t/svgelements/releases/tag/1.8.1) of `svgelements`. Issue #1760 is now fixed. So I removed the process that expands all elements referenced by `<use>`, and add the version limitation of `svgelements`.

3b1b:
Looks good to me.

If possible in the future (and I fully recognize the hypocrisy in this), it's better to split up conceptually distinct changes into multiple different PRs, e.g. the introduction of templates vs. the various issues fixed.

PR #1865, opened 2022-09-14
Added Auto Zoom functionality to scene.py

HadiSDev:
## Motivation

I wanted to add the Auto Zoom functionality back to manim as it is very useful :) And saves a lot of time
## Proposed changes

- scene.py
I Basically copied the old implementation 
## Test

**Code**:
```
class SquareToCircle(Scene):
    def construct(self):
        circle = Circle()
        circle.set_fill(BLUE, opacity=0.5)
        circle.set_stroke(BLUE_E, width=4)
        square = Square()

        self.play(ShowCreation(square))
        self.wait()
        self.play(ReplacementTransform(square, circle))
        self.wait()
        square = Square().move_to(LEFT)
        self.play(ShowCreation(square))
        self.play(self.auto_zoom([square]))
```
**Result**:
image

### DIFF
--- a/manimlib/scene/scene.py
+++ b/manimlib/scene/scene.py
@@ -304,6 +304,88 @@ class Scene(object):
             for sm in mob.get_family()
         ])
 
+    def auto_zoom(self,
+            mobjects: list[Mobject],
+            margin: float = 0,
+            only_mobjects_in_frame: bool = False,
+            animate: bool = True,
+    ):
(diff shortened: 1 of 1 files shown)


3b1b:
Thanks for contributing! A few comments:

- I think a lot of the implementation computes what you would get with the simpler line Group(*mobjects).get_bounding_box()
- I wonder if it would be better to just make this a method of CameraFrame, so that one might call self.play(frame.animate.zoom_to(square)), or something like that.

PR #1923, opened 2022-12-03
correcting  python version

louzanizineddine:
correcting the python version on the readme file.


TonyCrane:
Manim is not available on python 2.

Issue #1844, opened 2022-07-26
Is the function is_off_screen() implemented correctly?


yc-cui:
### Describe the bug

The function `is_off_screen()` , which determines whether or not some `Mobject` is out of frame, is weird.
Should `get_left()` and `get_right()`,  `get_bottom()`  and  `get_top()` reverse position?
**Code**:

    def is_off_screen(self):
        if self.get_left()[0] > FRAME_X_RADIUS:
            return True
        if self.get_right()[0] < -FRAME_X_RADIUS:
            return True
        if self.get_bottom()[1] > FRAME_Y_RADIUS:
            return True
        if self.get_top()[1] < -FRAME_Y_RADIUS:
            return True
        return False

**Wrong display or Error traceback**:

None
### Additional context

None

TonyCrane:
The goal of this method is to determine if the object is **completely** off the screen, instead of returning True whenever just a part of it is off screen.

PR #1924, opened 2022-12-08
Fix bug: Issue#1844 Modified is_off_screen() function

MashyBasker:
## Motivation
To fix the issue https://github.com/3b1b/manim/issues/1844#issue-1317730247
## Proposed changes
- Changed the comparison conditions to check if object is outside the screen 

**Code**:
```py
def is_off_screen(self):
        if self.get_left()[0] < -FRAME_X_RADIUS:
            return True
        if self.get_right()[0] > FRAME_X_RADIUS:
            return True
        if self.get_bottom()[1] < -FRAME_Y_RADIUS:
            return True
        if self.get_top()[1] > FRAME_Y_RADIUS:
            return True
        return False
```

### DIFF
--- a/manimlib/mobject/mobject.py
+++ b/manimlib/mobject/mobject.py
@@ -1035,13 +1035,13 @@ class Mobject(object):
         return self
 
     def is_off_screen(self):
-        if self.get_left()[0] > FRAME_X_RADIUS:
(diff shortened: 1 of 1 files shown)


TonyCrane:
See https://github.com/3b1b/manim/issues/1844#issuecomment-1345285896

PR #1932, opened 2022-12-17
Kill config

3b1b:
- Removing usage of CONFIG dicts, handling that state either in __init__ parameters are class variables, depending on which makes more sense to the user.
- Add clearer type hints throughout
- A few bits of refactoring and cleanup throughout the library
- Removing old or unused constructs which were specific to 3b1b videos, or old practices.


3b1b:
@TonyCrane Just a heads up, I think this may affect some of your documentation. I'm guessing it makes sense to simply delete the page of the docs talking about CONFIG dicts?

TonyCrane:
> @TonyCrane Just a heads up, I think this may affect some of your documentation. I'm guessing it makes sense to simply delete the page of the docs talking about CONFIG dicts?

Yes, that page can be deleted.

By the way, after a while I plan to check the documentation completely, update the changelog, and prepare to release 1.7.0. What about your opinion?

manimlib/mobject/number_line.py
```diff
@@ -10,77 +10,83 @@
...
-        super().__init__(self.x_min * RIGHT, self.x_max * RIGHT, **kwargs)
-        if self.width:
-            self.set_width(self.width)
-            self.unit_size = self.get_unit_size()
+        numbers_with_elongated_ticks: list[float] = [],
+        include_numbers: bool = False,
+        line_to_number_direction: Vect3 = DOWN,
+        line_to_number_buff: float = MED_SMALL_BUFF,
+        include_tip: bool = False,
+        tip_config: dict = dict(
```
YishiMichael, line 42: I noticed there are many places using dicts (mutable) as default arguments. It could be better to convert them into None and set the default values accordingly.

3b1b: I agree. This initial pass was mainly geared at converting CONFIG state to __init__ parameters, fairly directly, but I'll do another sweep for things like that.

3b1b:
> By the way, after a while I plan to check the documentation completely, update the changelog, and prepare to release 1.7.0. What about your opinion?

That sounds good to me, though I might give a little breathing room between this PR and any new release to give time for unexpected issues to make themselves known.

PR #1940, opened 2022-12-20
Single svg write

3b1b:
Minimal changes to illustrate behavior of StringMobject with a single svg write

### DIFF
--- a/manimlib/config.py
+++ b/manimlib/config.py
@@ -1,6 +1,8 @@
+from __future__ import annotations
+
 import argparse
+from argparse import Namespace
 import colour
-from contextlib import contextmanager
 import importlib
 import inspect
 import os
@@ -12,6 +14,10 @@ from manimlib.logger import log
 from manimlib.utils.dict_ops import merge_dicts_recursively
 from manimlib.utils.init_config import init_customization
 from manimlib.constants import FRAME_HEIGHT
+
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    Module = importlib.util.types.ModuleType
 
 
 __config_file__ = "custom_config.yml"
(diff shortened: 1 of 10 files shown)


3b1b:
@YishiMichael, I'm trying to better understand why MTex will write two svgs rather than just one. I realize the intent is to have one which doesn't have all the color labeling and to use the color-labeled one purely to generate spans etc., but why not just create the color-labeled one alone and use that? The style can (and by default will) be overwritten based on the fill_color/stroke_color/etc.

I'm guessing there are cases I'm just not thinking of. The code in this PR is just the minimum possible change to demonstrate what it would look like to only use the color-labeled svg, not a serious suggestion for how to change it. Do you have examples in mind where this would cause problems?

YishiMichael:
I've thought of cases where labelling will change the **shape** of svg. The mechanism of labelling a substring is done via replacing `some random substring` with `{{\color[RGB]{0,0,1}some random substring]}}`. The following two common cases come to my mind.
1. `a_1^2` -> `{{a_1}}^2` (the exponent part will be shifted to right)
2. `a + b` -> `a {{+}} b` (the spaces around the plus sign will be stripped)

To be honest, I'm not sure whether labelling won't change the shape in most use cases. I've also thought of methods to judge whether to create a labelled svg from substrings given, ending up being lost in really complex code. In earlier versions of MTex labelled svgs are created first, and a checklist is used to judge whether original svgs need to be created. This can be referred as an alternative of refactory, and the question is there may be far more cases where braces will change the shape.

For the sake of efficiency and simplicity, another approach (personally recommended) is to simply add a new parameter, leaving the chance to judge whether to generate the original svg to the user.

3b1b:
> For the sake of efficiency and simplicity, another approach (personally recommended) is to simply add a new parameter, leaving the chance to judge whether to generate the original svg to the user.

This makes a lot of sense to me. Later I'll update this PR with a concrete suggestion for how that might look. My guess is that in 99% of cases it's fine to just use the color-labeled generation. For those 1% of cases where something ends up a little misaligned, adding an option to ensure alignment using a second call to probably wouldn't be too bad.

What I'd like to do is replace the current Tex and MTex (or rather, deprecate the current Tex and rename MTex to Tex). It will definitely feel nicest if the default case involves only a single call to 'latex'

PR #1981, opened 2023-01-31
Add Self type

3b1b:
Add [Self](https://peps.python.org/pep-0673/) as the return type for many Mobject methods

### DIFF
--- a/manimlib/mobject/changing.py
+++ b/manimlib/mobject/changing.py
@@ -11,7 +11,7 @@ from manimlib.utils.rate_functions import smooth
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Callable, List, Iterable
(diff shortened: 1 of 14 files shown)


TonyCrane:
However, `Self` is new in python 3.11 (as stated in PEP 673 is for version 3.11). So after adding `Self` in this PR, manim will not work with 3.10 and earlier, and will raise the following exception.

```text
ImportError: cannot import name 'Self' from 'typing'
```

A workaround is to use the [typing-extensions](https://github.com/python/typing_extensions) package.

For example, use the try-except statement to catch the error of `from typing import Self`, and use `from typing_extensions import Self` if it raises an ImportError. This ensures that manim will still work with version 3.7+.

In addition to this, you need to add a `typing-extensions` dependency.

```text
typing-extensions ; python_version < "3.11"
```

PR #2044, opened 2023-07-16
added information for M1 Users

eivaremir:
## Motivation

Had a really bad time trying to make it work for M1 Apple 

## Proposed changes
- Added a small instruction to readme file

## Test

**Code**:
manimgl in the terminal
**Result**:
executed correctly

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,13 @@ manim-render example_scenes.py OpeningManimExample
     ```sh
     brew install ffmpeg mactex
     ```
+
+2. If you are using M1, install Cairo. 
+    ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,8 +67,13 @@ manim-render example_scenes.py OpeningManimExample
     ```sh
     brew install ffmpeg mactex
     ```
+
+2. If you are using M1, install Cairo.
```
pyxelr, line 71: I think it should be more generalized:

```suggestion
2. If you are using an ARM-based processor, install Cairo. 
```

PR #2061, opened 2023-09-03
Error effect when rendering the same ImageMobject

floatwoozy:
## Motivation
Triggered when rendering the same texture consecutively.
Mobject default render mode is TRIANGLE_STRIP, which will outputs extra triangles;
Use TRIANGLES mode and add vertices to fix this.

## Proposed changes
- image_mobject.py

## Test
**Code**:
```python
class Test(Scene):
    def construct(self):
        image = ImageMobject("res/bili.png").scale(0.2)
        a = Group(*[image.copy() for _ in range(25)]).arrange_in_grid(5, 5)
        self.add(a)
```

**Result**:
Test
Test_f

### DIFF
--- a/manimlib/mobject/types/image_mobject.py
+++ b/manimlib/mobject/types/image_mobject.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import numpy as np
 from PIL import Image
+from moderngl import TRIANGLES
 
 from manimlib.constants import DL, DR, UL, UR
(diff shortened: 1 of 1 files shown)


3b1b:
Thanks!

3b1b:
Also, for what it's worth, the second line could be accomplished by writing "a = image.get_grid(5, 5)"

PR #2147, opened 2024-07-23
fix(animation): `time_span` doesn't work for mobjects with submobjects

osMrPigHead:
## Motivation / 修改原因

`time_span` doesn't work for animations of mobjects with submobjects, such as `DashedLine`.
像 `DashedLine` 这样含有多个子物件的图形的动画无法正确处理 `time_span`。

## Proposed changes / 修改内容

- add method `time_spanned_alpha` in `Animation` to calculate the animation progress instead of changing `rate_func` when `begin` is called
  在 `Animation` 类中添加了计算动画进度的方法 `time_spanned_alpha`，专门用来计算存在 `time_span` 时的动画进度，而不在调用 `begin` 时改变 `rate_func`
- since `Rotating` also uses `rate_func` to decide the animation progress in `interpolate_mobject`, changes of `rotation.py` are made to correspond
  由于 `Rotating` 类的 `interpolate_mobject` 也是用 `rate_func` 计算动画进度，为了保持一致还修改了 `rotation.py`

## Test / 测试

**Code / 代码**:
```python
from manimlib import *


class TestScene(Scene):
    def construct(self) -> None:
        self.play(ShowCreation(Line((-1, 1, 0), (-1, -1, 0)), time_span=(0, 2)),
                  ShowCreation(DashedLine((1, 1, 0), (1, -1, 0)), time_span=(0.7, 2.7)))
        self.wait()
```
We hope the dashed line on the right side start to show later than the solid line on the left side.
右侧动画应比左侧晚开始。

**Result / 结果**:
After this change: (obviously the right one start later)
修改后：（右侧动画开始更晚）

Before this change: (two animations seem to start at the same time)
修改前：（两侧动画几乎同时开始）

### DIFF
--- a/manimlib/animation/animation.py
+++ b/manimlib/animation/animation.py
@@ -62,9 +62,6 @@ class Animation(object):
         if self.time_span is not None:
             start, end = self.time_span
             self.run_time = max(end, self.run_time)
-            self.rate_func = squish_rate_func(
-                self.rate_func, start / self.run_time, end / self.run_time,
-            )
         self.mobject.set_animating_status(True)
         self.starting_mobject = self.create_starting_mobject()
         if self.suspend_mobject_updating:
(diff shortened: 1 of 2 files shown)


3b1b:
Good fix, thanks!

PR #2233, opened 2024-10-24
Properly check modifier keys.

mitkonikov:
## Motivation

Modifier keys are given by Pyglet with bitmasks, however the checks were done by equality comparison. Also, the checks were only done for Mac users (COMMAND key) and not the CTRL command. In case the user had the NUM_LOCK or CAPS_LOCK or any other modifier key turned on, then the checks completely failed and nothing was interactive.

## Proposed changes
- Change the checks with the `&` operator.
- Add both `COMMAND` and `CTRL` checks
- Properly get the modifier constants from the Pyglet library instead of hardcoding them
- Remove the old hardcoded now unused modifier constants

## Test

**Code**:

```py
from manimlib import *
import numpy as np

class TestScene(InteractiveScene):
    def construct(self):
        ## Interactions
        circle = Circle(radius = 2)
        self.play(DrawBorderThenFill(circle))
```

Run the simple interactive scene and test if all commands work properly.

### DIFF
--- a/manimlib/constants.py
+++ b/manimlib/constants.py
@@ -92,10 +92,6 @@ COMMAND_SYMBOL: int = 65517
 DELETE_SYMBOL: int = 65288
 ARROW_SYMBOLS: list[int] = list(range(65361, 65365))
 
-SHIFT_MODIFIER: int = 1
-CTRL_MODIFIER: int = 2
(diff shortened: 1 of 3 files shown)


3b1b:
Thanks, this is notably cleaner.

PR #2241, opened 2024-11-10
Add `focus()` command

Splines:
## Motivation

When working in the interactive mode, the OpenGL window might be hidden behind other programs. Therefore, we introduce a new `focus()` command here to focus the window. This could also be very beneficial to our [`Manim Notebook`](https://github.com/Manim-Notebook/manim-notebook) VSCode extension (in construction) allowing to automatically focus when previewing a Manim Cell.

## Proposed changes
- Add a `focus()` method to the scene.
- Implement the focus mechanism in the window.

Unfortunately, the pyglet [`activate()`](https://pyglet.readthedocs.io/en/latest/modules/window.html#pyglet.window.Window.activate) method did not work for me (using WSL) while the proposed solution works, but produces a small flickering and also offsets the window by a small amount in the upper left direction. Maybe you know a better approach?

## Test

Just take any Manim code and get into the interactive mode.

```py
from manimlib import *


class ReloadTest(Scene):
    def construct(self):
        ## Circle creation
        circle = Circle()
        self.play(ShowCreation(circle))

        ## Circle movement
        self.play(circle.animate.shift(2 * UP + 2 * RIGHT))
```

```
manimgl ./reload_sample.py ReloadTest -se 6
```

Then issue the `reload()` command.

### DIFF
--- a/manimlib/scene/scene.py
+++ b/manimlib/scene/scene.py
@@ -234,6 +234,7 @@ class Scene(object):
             add=self.add,
             remove=self.remove,
             clear=self.clear,
+            focus=self.focus,
             save_state=self.save_state,
             undo=self.undo,
             redo=self.redo,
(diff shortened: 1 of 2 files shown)


3b1b:
Thanks for the contribution!

PR #2251, opened 2024-11-27
Rereload session modules during `reload()`

Splines:
## 🎈 Motivation

In #2240, I've added a `reload()` command that will reuse the OpenGL window to speed up reloading (instead of having to quit the active ManimGL Python interpreter session, then starting it again). Unfortunately, I've missed out on the following point [from this comment](https://github.com/3b1b/manim/pull/2240#issuecomment-2501620383).

> In fact, for **code changes outside the file of the scene**, from what I can tell the reload call does not incorporate those, whereas the command+A, command+R on that file would.

While in theory the `reload()` command will perform every single step that we also do when entering ManimGL for the very first time (except for parsing arg params), in practice we observe some module caching which this PR addresses.

See also the test example down below. Before this PR, any changes to file `b.py` won't be reflected, even when using the `reload()` command.

## 💻 Proposed changes

**All changes discussed here only affect Manim when the `reload()` command is called, everything else remains the same** (except for the additional `import sys` which really shouldn't be too much overhead).

This is a bit delicate since we have to distinguish between
- modules that we rely on in ManimGL itself before spawning an [`InteractiveShellEmbed`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.terminal.embed.html#IPython.terminal.embed.InteractiveShellEmbed)
- modules that the user loads in their Python files that define the scene

Both of these modules are somehow managed globally by the Python interpreter. What I do in this PR is
- before running the first scene, store a set of modules that were loaded so far
- when the user triggers the `reload()` command, we "unload" all modules that were added to this set in the meantime

Therefore, those user-defined "session modules" will be reloaded again when the next scene starts by means of [this already existing Manim code line](https://github.com/Splines/manim-fork/blob/626dd1fe3343de0da366f9fd827fbc88f7b99024/manimlib/config.py#L213).

## 🧪 Test

```py
# file a.py
import sys

sys.path.append(".")
from manimlib import *

from b import B

class AScene(Scene):
    def construct(self):
        ## Starting
        print("Hello from AScene")

        ## B
        b = B()
        print(b.answer_to_life())
```

```py
# file b.py
class B:
    def answer_to_life(self):
        return 10
```

```bash
# Line 13 is where the comment "##B" is
manimgl "/path/to/a.py" AScene -se 13
```

- Preview by starting from the second comment (`##B`) and get `10` printed to the console.
- Change `10` to `42` in `b.py` and save the file
- Preview again, still see `10` in the console.
- Do a `reload(13)`
- Preview again, now see the `42` in the console 🎉

### DIFF
--- a/manimlib/reload_manager.py
+++ b/manimlib/reload_manager.py
@@ -1,17 +1,7 @@
+import sys
+
 from typing import Any
 from IPython.terminal.embed import KillEmbedded
-
-
-def restart_entirely():
-    """
-    Restarts the Python interpreter. The whole state will be lost as this is
-    equivalent to exiting the terminal, then issuing the ManimGL command again.
-    """
-    import os
-    import sys
-    print("Restarting...")
-    python = sys.executable
-    os.execv(python, [python] + sys.argv)
 
 
 class ReloadManager:
@@ -37,11 +27,29 @@ class ReloadManager:
         """
         self.start_at_line = start_at_line
 
+    def delete_newly_loaded_modules(self, initial_modules):
+        """
+        Deletes the modules that were loaded during the last scene run.
+
+        These are the modules that were not present in the initial set of
+        modules that were loaded when ManimGL was started. I.e. we delete
+        modules that the user imported in their own Python files such that
+        any changes to these files are reflected in the next scene run.
+        """
+        current_modules = set(sys.modules.keys())
+        session_modules = current_modules - initial_modules
+
+        for module_name in session_modules:
+            if module_name in sys.modules:
+                del sys.modules[module_name]
+
     def run(self):
         """
         Runs the scenes in a loop and detects when a scene reload is requested.
         """
         while True:
+            initial_modules = set(sys.modules.keys())
+
             try:
                 # blocking call since a scene will init an IPython shell()
                 self.retrieve_scenes_and_run(self.start_at_line)
@@ -56,7 +64,7 @@ class ReloadManager:
 
                 self.scenes = []
 
-                restart_entirely()
+                self.delete_newly_loaded_modules(initial_modules)
 
             except KeyboardInterrupt:
                 break
--- a/manimlib/reload_manager.py
+++ b/manimlib/reload_manager.py
@@ -2,6 +2,18 @@ from typing import Any
 from IPython.terminal.embed import KillEmbedded
 
 
+def restart_entirely():
+    """
+    Restarts the Python interpreter. The whole state will be lost as this is
+    equivalent to exiting the terminal, then issuing the ManimGL command again.
+    """
+    import os
+    import sys
+    print("Restarting...")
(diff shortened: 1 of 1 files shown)


3b1b:
Trying this, I get the following error. I believe this is specifically about a failure within the screeninfo module, but I'm guessing it's a symptom of there being broader potential unintended consequences with deleting previously loaded modules like this.


```
UnboundLocalError                         Traceback (most recent call last)               
File /opt/miniconda3/bin/manimgl:8                                                        
      6 if __name__ == '__main__':                                  
      7     sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
----> 8     sys.exit(main())                                                                                                             
                                                                                                                                         
File ~/cs/manim/manimlib/__main__.py:26, in main()                                                                                       
     23     return                                                                                                                       
     25 reload_manager.args = args
---> 26 reload_manager.run()

File ~/cs/manim/manimlib/reload_manager.py:59, in ReloadManager.run(self)
     55 initial_modules = set(sys.modules.keys())
     57 try:                
     58     # blocking call since a scene will init an IPython shell()
---> 59     self.retrieve_scenes_and_run(self.start_at_line)
     60     return
     61 except KillEmbedded:
     62     # Requested via the `exit_raise` IPython runline magic                                                
     63     # by means of our scene.reload() command
                                                           
File ~/cs/manim/manimlib/reload_manager.py:90, in ReloadManager.retrieve_scenes_and_run(self, overwrite_start_at_line)
     87     self.args.embed = str(overwrite_start_at_line)
     89 # Args to Config                         
---> 90 config = manimlib.config.get_configuration(self.args)
     91 if self.window:
     92     config["existing_window"] = self.window  # see scene initialization
                                                                                                                                         
File ~/cs/manim/manimlib/config.py:503, in get_configuration(args)
    501 custom_config = get_custom_config()
    502 camera_config = get_camera_config(args, custom_config)
--> 503 window_config = get_window_config(args, custom_config, camera_config)
    504 start, end = get_animations_numbers(args)
    506 return {
    507     "module": get_scene_module(args),
    508     "scene_names": args.scene_names,
   (...)
    523     "embed_error_sound": custom_config["embed_error_sound"],
    524 }

File ~/cs/manim/manimlib/config.py:440, in get_window_config(args, custom_config, camera_config)
    438     pass
    439 mon_index = custom_config["window_monitor"]
--> 440 monitor = monitors[min(mon_index, len(monitors) - 1)]
    441 aspect_ratio = camera_config["pixel_width"] / camera_config["pixel_height"]
    442 window_width = monitor.width

UnboundLocalError: cannot access local variable 'monitors' where it is not associated with a value
```

Splines:
Oh, thanks for noting this, I am now working on a new solution and will file in a new PR in a few days.

PR #2252, opened 2024-11-28
Add autoreload mechanism

Splines:
## 🎈 Motivation

IPython offers an `%autoreload all` magic command:

> `autoreload` reloads modules automatically before entering the execution of code typed at the IPython prompt.

This makes it possible for Manim users to make changes in imported files and **see those changes reflected automatically without having to reload manually** (by means of quitting and restarting the terminal, or by means of the new `reload()` command from #2251). Therefore, this PR improves the interactive experience with Manim fostering playful change of parameters/functions etc.

## 💻 Proposed changes

- Add a new `--autoreload` CLI flag.
- If that flag is set, call the respective magic commands before the [`InteractiveShellEmbed`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.terminal.embed.html#IPython.terminal.embed.InteractiveShellEmbed) shell is started.

Note that `autoreload` should probably not be activated by default because it will slow down the performance since modules will be reloaded automatically and maybe even in cases where it wouldn't be necessary. Furthermore, we should be aware of the [caveats](https://ipython.readthedocs.io/en/stable/config/extensions/autoreload.html#caveats) that come with `autoreload`. This is why I went for exposing it as flag that can be activated if users like to.

## 🧪 Test

```a.py
# file a.py
import sys

sys.path.append(".")
from manimlib import *

from b import B

class AScene(Scene):
    def construct(self):
        ## Starting
        print("Hello from AScene")

        ## B
        b = B()
        print(b.answer_to_life())
```

```py
# file b.py
class B:
    def answer_to_life(self):
        return 10
```

```bash
# Line 13 is where the comment "## B" is
manimgl "/path/to/a.py" AScene -se 13 --autoreload
```

- Preview by starting from the second comment (`## B`) and get `10` printed to the console.
- Change `10` to `42` in `b.py` and save the file
- Preview again and see `42` in the console, without having to reload manually 🎉

> [!tip]
> Unfortunately, classes inside the same file won't be autoreloaded, just modules that are imported. E.g. see the evolute test example provided in #2240. So when defining custom classes that don't just host Manim rendering logic like `self.play()` etc., it might be worth to outsource such parts to another file when we want to work on them and make lots of changes. Otherwise, starting with #2251, the `reload()` command such also work in such cases.

If you want to reload classes from other modules, I just found [this post](https://anthe.sevenants.net/post/reload-classes-ipython).

## 📜 Documentation
I've added the CLI flag `--autoreload` to the docs.

<details>

<summary>I also tried to build the docs but didn't succeed</summary>

I wanted to test this by following [your instructions from here](https://3b1b.github.io/manim/development/contributing.html#how-to-build-this-documentation). Unfortunately, I stumbled into these issues:

### Jinja
I first got [this error](https://github.com/sphinx-doc/sphinx/issues/10291) and therefore downgraded Jinja:

```bash
$ pip install "jinja2<3.1"

> Successfully uninstalled Jinja2-3.1.4
> Successfully installed jinja2-3.0.3
```

### Sphinx

Then I tried to run `make html` again which failed with:

```py
WARNING: while setting up extension sphinx.addnodes: node class 'meta' is already registered, its visitors will be overridden

Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
```

Realizing that `Sphinx==3.0.3` is specified in `docs/requirements.txt`, I quit this pursuit to build the docs as I was not even sure it would work on your end...
</details>


Splines:
Nice, the new scene_embed makes it a lot easier to get a grasp on what is going on! I've now made my PR ready for the new changes.

PR #2257, opened 2024-12-04
Reload user-defined modules during `reload()`

Splines:
#2251 didn't work, so here is a new attempt :)

## 🎈 Motivation

In #2240, I've added a `reload()` command that will reuse the OpenGL window to speed up reloading (instead of having to quit the active ManimGL Python interpreter session, then starting it again). Unfortunately, I've missed out on the following point [from this comment](https://github.com/3b1b/manim/pull/2240#issuecomment-2501620383).

> In fact, for **code changes outside the file of the scene**, from what I can tell the reload call does not incorporate those, whereas the command+A, command+R on that file would.

While in theory the `reload()` command will perform every single step that we also do when entering ManimGL for the very first time (except for parsing arg params), in practice we observe some module caching which this PR addresses.

> [!tip]
> With regards to the main use case: yes, one could just select the changed code and run it directly in the terminal to update it. But this is not always convenient, e.g. imagine you want to change the code at multiple occurrences (maybe even spread across different files) and see the effects. E.g. consider also my [evolute example here](https://github.com/3b1b/manim/pull/2240#issuecomment-2498946548). Or the case where you are a Manim developer and want to see the effects directly when you changed multiple files deep down at the heart of Manim (set `IGNORE_MANIMLIB_MODULES = False` in this case).
> 
> The main idea behind `reload()` is that in situations where a `checkpoint_paste()` might not work or is cumbersome, instead of having to `Ctrl+C` the active session, then reloading everything again, we have a `reload(<optional_line_number>)` command as introduced in #2240. To the user, the final behavior should look exactly like `Ctrl+C` then spinning up again, with the benefit of the OpenGL window staying open all the time and a lot faster reload times (like milliseconds vs. multiple seconds).

## 💻 Proposed changes

We introduce a new `ModuleLoader`. If the user doesn't execute `reload()`, the `get_module()` method of that loader is exactly the same as beforehand with the main part being:

```py
module_name = file_name.replace(os.sep, ".").replace(".py", "")
spec = importlib.util.spec_from_file_location(module_name, file_name)
module = importlib.util.module_from_spec(spec)
spec.loader.exec_module(module)
```

as inspired [by the Python docs](https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly). So good news, we should not break anything when `reload()` is not executed.

Now to the interesting case _with_ `reload()`. A first attempt was made in #2251 by deleting module names from the `sys.modules` to force them being reloaded. This didn't end up well and resulted in error messages for third-party libraries.

In this PR, we use the proper way of **`importlib.reload(module)`**. But it's unfortunately not as simple as that, so here's an overview of what we do (details in docstrings):
- whenever we are in the `reload()` case...
- Execute `spec.loader.exec_module(module)` with a replaced `__import__` function that tracks those imports, such that we know what modules the user imported in their files.
- Out of those gathered modules, reload those that are "user-defined", e.g. when custom files are imported, but NOT built-in Python system modules like `sys`, `time` etc. and NOT external 3rd-party libraries like `numpy` etc.
- To achieve this reloading across multiple levels, we implement a recursive `deep_reload()`. I've tried with [`IPython.lib.deepreload`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.lib.deepreload.html) but this doesn't work since also system modules and 3rd-party modules will be imported (and those should not be reloaded).
- Note: The `IGNORE_MANIMLIB_MODULES` controls whether manimlib modules should be ignored. I've set this to `True` by default since most users won't make any changes to the Manim code itself. Set it to `False` and the `reload()` will take a bit longer to reload everything (still a lot quicker than killing ManimGL altogether and reinitializing it).

## 🧪 Test

<details>

<summary>See test files</summary>

```py
# file a.py
import sys

sys.path.append(".")
from manimlib import *
import numpy as np  # just to have another dependency to test the import

from b import BAndC


class D:
    def answer_d(self):
        return "Hello from D"


class AScene(Scene):
    def construct(self):
        ## Starting
        print("Hello from A")

        ## B
        b_and_c = BAndC()
        print(b_and_c.answer_b())

        ## C
        print(b_and_c.answer_c())

        ## D
        d = D()
        print(d.answer_d())
```

```py
# file b.py
from c import C

class BAndC:
    def __init__(self):
        self.c = C()

    def answer_b(self):
        return "Hello from B"

    def answer_c(self):
        return self.c.answer_c()
```

```py
# file c.py
class C:
    def answer_c(self):
        return "Hello from C"
```

</details>

```bash
manimgl "/path/to/a.py" AScene -se 29
```

- Run and see the respective `Hello from X` statements printed to the console.
- Change any string (a, b, c or d) and save the file.
- Preview via `checkpoint_paste()`: still see the old strings printed.
- `reload()`
- Now see the updated statements in the console and also debug messages of which modules were reloaded 🎉 (e.g. here a, c and b in that order)

image


Splines:
🎬 And here two videos to showcase this.

First the example from the PR comment:

https://github.com/user-attachments/assets/ccc154ef-358a-45c1-957b-4dbc7bf41051

An then how this could be beneficial for Manim devs by setting `IGNORE_MANIMLIB_MODULES = False`.

https://github.com/user-attachments/assets/9ce45145-5289-4fc8-9bc1-678a44087184

3b1b:
This works alright for me now, nice!

I appreciate the contribution, and am happy to merge, though I do feel like I should express some wariness around adding complexity without a strong need. In this case, quitting and restarting is perhaps a little bothersome—a second or two of of waiting in an otherwise smooth workflow can be—but it's right on the edge.

The module_loader.py functionality is flirting with functionality that feels best suited for python itself to manage: Tracking which modules are built in, making a wrapper around builtins.__import__, etc. I understand how given what you want to do, it makes sense to have all this, and you've done it very cleanly. But in general, I feel like manim has a lot of unnecessary complexity and hackiness in places, and it'd be nice to mitigate that or nip it in the bud when possible.

Splines:
Thanks a lot for reviewing and merging, this will allow me to include this in the Manim Notebook extension via https://github.com/Manim-Notebook/manim-notebook/pull/95

> The module_loader.py functionality is flirting with functionality that feels best suited for python itself to manage: [...]

Yes, indeed, it did indeed feel "hacky" at times and I was a bit disappointed that I couldn't find an easier way to do all this and that I had to dive that deeply into the Python module mechanism. I strongly hope that this will remain my last "hacky" contribution to Manim. <sub>I also hope that this PR doesn't introduce any unwanted side-effects down the line; feel free to ping me in an issue should this be the case somewhere in the future. But I feel pretty optimistic that it won't happen given that the code is really just executed when `reload()` is called.</sub>

> In this case, quitting and restarting is perhaps a little bothersome—a second or two of of waiting in an otherwise smooth workflow can be—but it's right on the edge.

Fair enough. Where `reload()` might shine is when you modify multiple places of code at the same time. In this case, you'd have to select all occurrences and run them one after the other (while maybe even considering the order of statements). Another case: when you're working in a bigger class, you are forced to always select the whole class, which in my eyes is tedious ;) That said, I do totally understand the wariness and try to avoid such PRs in the future.

PR #2259, opened 2024-12-05
Cleaner local caching of Tex/Text data, and partially cleaned up configuration

3b1b:
Previously, in generating Tex and Text mobjects, manim would write to and read from various svg files on disk, to prevent the need to regenerate that data on each run, which is especially relevant in rendering LaTeX.

With this PR, we instead use the cache functionality from the disk_cache library to locally cache the relevant svg strings. This makes it easier to consolidate code, and to clear the cache or limit how much disk space it uses.

This also involved refactoring SVGMobject to optionally take in an svg_string, and to update Tex and Text accordingly.

---

Separately, this PR also includes a fair bit of cleaning up for within config.py.  It remains somewhat messy, in particular it's very ambiguous what "config" exactly refers to throughout, but this is an attempt to simply how the default_config and custom_config yaml files are loaded, and how they interface with the CLI arguments.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,6 @@
+appdirs
 colour
+diskcache
 ipython>=8.18.0
 isosurfaces
 fontTools
@@ -21,6 +23,7 @@ screeninfo
 skia-pathops
 svgelements>=1.8.1
 sympy
+tempfile
 tqdm
 typing-extensions; python_version < "3.11"
 validators
--- a/manimlib/__init__.py
+++ b/manimlib/__init__.py
@@ -64,7 +64,6 @@ from manimlib.utils.bezier import *
 from manimlib.utils.cache import *
 from manimlib.utils.color import *
 from manimlib.utils.dict_ops import *
-from manimlib.utils.customization import *
 from manimlib.utils.debug import *
 from manimlib.utils.directories import *
 from manimlib.utils.file_ops import *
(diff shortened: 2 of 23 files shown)


requirements.txt
```diff
@@ -20,6 +22,7 @@ screeninfo
 skia-pathops
 svgelements>=1.8.1
 sympy
+tempfile
```
Splines, line 25: @3b1b This new line got me into some trouble and I couldn't install Manim anymore. I.e. `pip install -r requirements.txt` results in this error:

```
ERROR: Could not find a version that satisfies the requirement tempfile (from versions: none)
ERROR: No matching distribution found for tempfile
```

Maybe it's related to [this issue](https://stackoverflow.com/questions/44286628/could-not-find-a-version-that-satisfies-the-requirement-backports-tempfile-from)? For now, I removed this line locally to get Manim up and running again.

---

Note that originally I just ran `pip install -e .` as usual but that didn't install the new `appdirs` library and therefore Manim failed on startup. Isn't `pip install -e .` supposed to install the required dependencies?

Splines: Also when searching on [Pip](https://pypi.org/search/?q=tempfile), I can't seem to find any library that is called `tempfile`.

3b1b: I may have neglected to update the requirement.txt, I'll take a look shortly.

yoyonel: Hello :-)
You definitely need a CI for testing and showing for all users/dev how to install properly Manim and dependencies (OS/system dependencies).
I can (easily) do this for Linux (/Debian) OS, more difficult for me right now to write this workflow for Windows and MacOS platform (but others can replicate and adapt my attempt for Linux platform).
Btw very interesting project, i like it a lot :-)

PR #2264, opened 2024-12-10
Fixing a Cairo Bug on Windows OS

Varniex:
### Cairo Error

While creating the temporary svg file in `text_mobject.py`, Cairo throws an error `b'error while writing to output stream'` on Windows OS.

### Proposed Changes
To avoid that, `delete=False` should be passed in `NamedTemporaryFile` function of the `tempfile` package.

Also, minor cleaning up the imports.


3b1b:
Thanks for the catch.

PR #2266, opened 2024-12-11
Refactor config

3b1b:
Configuration was a bit of a mess, this is an attempt to clean it up and better clarify what's going on.

At a high level, the default_config.yml file should be where all configuration defaults are set, determining default parameters for various classes, constants, or other specifications a user may want to change. They can then create their own custom_config.yml file to update any of these to their own personal taste (for instance mine lives [here](https://github.com/3b1b/videos/blob/master/custom_config.yml)).

**Note, for anyone who had a custom_config.yml file, this update will require that it be updated to a new format**

The configuration from these files is loaded into a global dictionary, named manim_config, which is also updated based on the CLI arguments.

Previously, a lot of configuration that used CLI arguments, or these yml files, was a bit spread out an unsystematic. Here the aim is for configuration associated with specific classes to be clearly marked as such, and for code relevant to processing that configuration (e.g. for Window) to lie within the class itself.

---

Other updates
- The functionality of ReloadManger is replaced with a simpler run_scenes in __main__.py, and the updates necessary for the reloading are instead consolidated in reload_scene.
- Bug fix for the use of xelatex


Splines:
Just as a suggestion, it'd be amazing to **squash-merge** PRs as it's otherwise a bit difficult to keep track of what changed in Manim on the `master` branch since we get commits like `Change "style" in default config to "text"`, which are mixed with more "high-level" ones that aggregate multiple changes under one roof, like `Refactor config`, `Add reload() command` etc. When browsing through the `master` branch, it'd be great to just see those high-level commits and if I want to I can still look at the PR and see all the details.

PR #2268, opened 2024-12-12
Autoreload v2

3b1b:
Adapting https://github.com/3b1b/manim/pull/2252 by @Splines after recent refactor.


Splines:
Yes, it does, thanks! Might be worth to add the new CLI flag to the docs as I did in #2252 originally (failing the pipeline).

Issue #2271, opened 2024-12-12
Window focus will offset the window


Splines:
In #2241, I've added a new `focus()` command [via these lines](https://github.com/3b1b/manim/blob/744e695340c5691551f2650d99db71964cb292c7/manimlib/window.py#L98-L110). Then in [this commit](https://github.com/3b1b/manim/commit/185f6428264f25dcdbd4faeb6062dbc75d8b1ed9), 3b1b decided to `focus()` the window every time on initialization, i.e. also when the new `reload()` command is used.

The problem is that I had to use a workaround in that `focus()` command since the pyglet [`activate()`](https://pyglet.readthedocs.io/en/latest/modules/window.html#pyglet.window.Window.activate) method didn't have any effects for me. That's why I did a simple

```py
self._window.set_visible(False)
self._window.set_visible(True)
```

This consistently restores focus on the window, however it also offsets its position for me, which is quite annoying as the window might even jump around when reaching the corners of the screen.

> [!tip]
> I've opened a new issue on the pyglet repo [here](https://github.com/pyglet/pyglet/issues/1244) to track this issue.

Splines:
Thanks for the partial fix via #2274, at least the window doesn't jump around now when I use `reload()` (but of course doesn't get focus either, but this is a better state than beforehand). Hopefully, the team at pyglet will find a solution such that `focus()` works properly.

PR #2274, opened 2024-12-12
Move resizing out of Window.focus, and into Window.init_for_scene

3b1b:
Somewhat confusingly, the window viewport can get offset on initialization, and on reload, in a way that simply resizing the window (even to its current size) fixes.

Previously I had added this to Window.focus, while also calling focus in Window.init_for_scene.  Better is to not necessarily focus on init, and to put the sync line in there directly.  This should partially address the annoyance in https://github.com/3b1b/manim/issues/2271

### DIFF
--- a/manimlib/window.py
+++ b/manimlib/window.py
@@ -69,7 +69,8 @@ class Window(PygletWindow):
         mglw.activate_context(window=self, ctx=self.ctx)
         self.timer.start()
 
-        self.focus()
+        # This line seems to resync the viewport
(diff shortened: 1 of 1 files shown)


Splines:
What do you mean by "resyncing" in "resyncing the viewport" here?

3b1b:
The symptom is that the scene appears off center in the Window. I'm not 100% sure what's going on, if I was there would probably be a more principled fix than simply calling "resize" arbitrarily.

Issue #2275, opened 2024-12-12
Feature request: Autoreload for the same file


Splines:
In the last few weeks, 3b1b and me introduced these features to ease reloading:

- `reload()` command via #2240 
- `--autoreload` CLI flag via #2268

The `reload(x)` command is essentially like quitting Manim altogether then starting it again with the `-se` option set to line `x`, but without the hassle of doing this manually and waiting for the complete window to be restarted. Instead, we reuse the window. Note that only your user-defined modules will be reloaded, no external libraries like `numpy`, no Python in-built modules like `sys` or `time` and also no `manimlib` files will be reloaded (the latter can be adjusted via a config option though).

The `--autoreload` command line flag will activate the [`autoreload`](https://ipython.readthedocs.io/en/stable/config/extensions/autoreload.html#autoreload) IPython magic command. With this, your imported modules will be automatically reloaded for you such that you don't even have to execute `reload()` anymore.

---

The only missing piece (in my opinion), is a feature that enables auto-reloading also for classes/objects/etc. in the same file (and not only for imported files). E.g. consider this example

```py
from manimlib import *


class ClassInSameFile:
    def answer(self):
        return "Hello from class in same file"


class AScene(Scene):
    def construct(self):
        ## Starting
        print("Hello from A")

        ## Class in same file
        test = ClassInSameFile()
        print(test.answer())

```

The autoreload feature from IPython will not pick up on changes to the string `"Hello from class in same file"` . If you do a `reload(14)` and then the `checkpoint_paste()`, it will work fine, but for the lazy programmer, that is one `reload(14)` too much 😅 A current **workaround** is to outsource `ClassInSameFile` to another file. Then import that into the original file and autoreload will do its magic.

Note that even if autoreload was *perfect* in the above sense, `reload()` might still have its raison d'être in the case where autoreloading might not work (due to whatever reason, I'm sure there might exist some...).

Splines:
I've opened an issue for this on the IPython repository: https://github.com/ipython/ipython/issues/14617

Splines:
With #2278, we don't use the IPython magic command `%autoreload all` anymore, but the custom Python module loading introduced in #2257. That is, this issue is actually done, so I close it.

PR #2278, opened 2024-12-13
Have autoreload update shell namespace with reloaded module variables

3b1b:
In response to https://github.com/3b1b/manim/issues/2275

When --autoreload is set, during the interactive mode, before each cell execution the user's module will be reloaded.

@Splines, is this what you had in mind? At the moment I removed the calls to shell.magic("autoreload all") because it feels like this accomplishes the same, but it's very possible I'm missing something.

### DIFF
--- a/manimlib/scene/scene_embed.py
+++ b/manimlib/scene/scene_embed.py
@@ -143,8 +143,11 @@ class InteractiveSceneEmbed:
 
     def auto_reload(self):
         """Enables IPython autoreload for automatic reloading of modules."""
-        self.shell.magic("load_ext autoreload")
-        self.shell.magic("autoreload all")
+        def pre_cell_func(*args, **kwargs):
+            new_mod = ModuleLoader.get_module(self.shell.user_module.__file__, is_during_reload=True)
+            self.shell.user_ns.update(vars(new_mod))
+
+        self.shell.events.register("pre_run_cell", pre_cell_func)
 
     def checkpoint_paste(
         self,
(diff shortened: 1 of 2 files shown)


Splines:
Nice, this is indeed exactly what I had in mind, it works perfectly, thank you 🙏 And yes, I think we don't need the IPython magic anymore since this is doing what we need. Maybe there are some exotic cases where autoreload does something we don't do, but in my basic playing around with this feature I haven't found it yet.

One thing to note is that while _adding_ import statements on the fly works, _removing_ them doesn't have any effect and they are still cached. A simple `reload()` works in this case. Maybe we could delete the old module namespace before updating it to mitigate this? However, if there's no easy solution to this very edge case, I really don't mind and I can just as well use `reload()` once in a while 😅

Furthermore, note that the pipeline in #2268 failed such that the `--autoreload` CLI flag is not yet published to the docs [here](https://3b1b.github.io/manim/getting_started/configuration.html#all-supported-flags).

manimlib/scene/scene_embed.py
```diff
@@ -143,8 +143,11 @@ def reload_scene(self, embed_line: int | None = None) -> None:
 
     def auto_reload(self):
         """Enables IPython autoreload for automatic reloading of modules."""
```
Splines, line 145: might want to adjust the docstring to make clear we don't use the `%autoreload` IPython line magic and instead reload user modules with our custom logic

3b1b:
> Maybe we could delete the old module namespace before updating it to mitigate this? However, if there's no easy solution to this very edge case, I really don't mind and I can just as well use reload() once in a while

That would be tricky, and prone to producing hard-to-detect errors. If you set up a keyboard shortcut for calling reload with the line number where the cursor is, it's really much of a cost at all.

I am less familiar with the docs build system. If you had any interest in looking into a fix, I'd certainly be grateful.

PR #2279, opened 2024-12-15
Fixed typo in default_config.yml

loay-4:
Changed --uhd to --4k


Varniex:
Why would you do that?!

The command for 4k is `--uhd` in `config.py` file.

Varniex:
The command for 4k is --uhd in config.py file. So, there's no need for the change.

PR #2280, opened 2024-12-16
Automatically identify the class name based on the specified line number.

Varniex:
### Motivation
Whenever some user passes the command `-e <line_number>`, it feels redundant to specify the class name.

The closest class name (above the specified `<line_number>`) should automatically be identified by the program, and updates the value of `scene_names` key in the `run_config` dict.

### Proposed Changes
I couldn't find a better position to insert the code than in the `insert_embed_line_to_module` function. As it ensures that the `scene_names` updates only when the `e` flag is passed, and also the module lines can be used to determine the class name.

Also, resolving minor bugs in `text_mobject` and `string_mobject` files.

### DIFF
--- a/manimlib/default_config.yml
+++ b/manimlib/default_config.yml
@@ -5,13 +5,13 @@
 # you are running manim. For 3blue1brown, for instance, mind is
 # here: https://github.com/3b1b/videos/blob/master/custom_config.yml
 
-# Alternatively, you can create it whereever you like, and on running
+# Alternatively, you can create it where ever you like, and on running
 # manim, pass in `--config_file /path/to/custom/config/file.yml`
 
 directories:
   # Set this to true if you want the path to video files
   # to match the directory structure of the path to the
-  # sourcecode generating that video
+  # source code generating that video
   mirror_module_path: False
   # Manim may write to and read from the file system, e.g.
   # to render videos and to look for svg/png assets. This
@@ -44,7 +44,7 @@ window:
   # If not full screen, the default to give it half the screen width
   full_screen: False
   # Other optional specifications that override the above include:
-  # position: (500, 500)  # Specific position, in pixel coordiantes, for upper right corner
+  # position: (500, 500)  # Specific position, in pixel coordinates, for upper right corner
   # size: (1920, 1080)  # Specific size, in pixels
 camera:
   resolution: (1920, 1080)
--- a/manimlib/default_config.yml
+++ b/manimlib/default_config.yml
@@ -5,7 +5,7 @@
 # you are running manim. For 3blue1brown, for instance, mind is
 # here: https://github.com/3b1b/videos/blob/master/custom_config.yml
 
-# Alternatively, you can create it where ever you like, and on running
+# Alternatively, you can create it wherever you like, and on running
 # manim, pass in `--config_file /path/to/custom/config/file.yml`
 
 directories:
--- a/manimlib/extract_scene.py
+++ b/manimlib/extract_scene.py
@@ -12,6 +12,7 @@ from manimlib.scene.interactive_scene import InteractiveScene
 from manimlib.scene.scene import Scene
 
 from typing import TYPE_CHECKING
+
 if TYPE_CHECKING:
     Module = importlib.util.types.ModuleType
     from typing import Optional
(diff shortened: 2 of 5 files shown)


manimlib/extract_scene.py
```diff
@@ -142,35 +143,48 @@ def get_indent(code_lines: list[str], line_number: int) -> str:
...
     lines = inspect.getsource(module).splitlines()
+    line_number = run_config.embed_line
 
     # Add the relevant embed line to the code
     indent = get_indent(lines, line_number)
     lines.insert(line_number, indent + "self.embed()")
     new_code = "\n".join(lines)
 
+    # When the user executes the `-e <line_number>` command,
+    # it should automatically identifies the nearest class
```
Splines, line 162:
```suggestion
    # it should automatically identify the nearest class
```

manimlib/default_config.yml
```diff
@@ -5,13 +5,13 @@
 # you are running manim. For 3blue1brown, for instance, mind is
 # here: https://github.com/3b1b/videos/blob/master/custom_config.yml
 
-# Alternatively, you can create it whereever you like, and on running
+# Alternatively, you can create it where ever you like, and on running
```
Splines, line 8:
```suggestion
# Alternatively, you can create it wherever you like, and on running
```

See [the Cambridge dictionary](https://dictionary.cambridge.org/dictionary/english/wherever)

manimlib/extract_scene.py
```diff
@@ -142,35 +143,48 @@ def get_indent(code_lines: list[str], line_number: int) -> str:
...
+    # defined above `<line_number>` as 'scene_names'.
+    classes = list(filter(lambda line: line.startswith("class"), lines[:line_number]))
```
Splines, line 164: What about classes that are defined as nested class, i.e. inside some other classes? They would have an indentation so `line.startswith("class")` would return `False`.

Varniex: I guess the Scene Class (like: `class SomeName(Scene)`) would not be the nested one. So, I think in that case as well, it works.

manimlib/default_config.yml
Splines: this was marked as resolved although it was not taken into account

manimlib/extract_scene.py
Splines: this was marked as resolved although it was not taken into account

Splines (changes requested):

manimlib/default_config.yml
Varniex: Thanks!

manimlib/extract_scene.py
Varniex: It's done!

3b1b:
There are some cases where the scene that has the embed line is not the same as the one I'm running. For example, suppose you define some SceneA with a certain construct. Then you define SceneB as a subclass, which doesn't have its own construct method (so it effectively uses that of SceneA), but makes other changes to the class variables, methods, etc. 
It may be that you want to drop an embed line in the construct method of SceneA, but to invoke SceneB.

The desired behavior is that if you _do_ pass in a scene names, like `manimgl file_name.py SceneB -se 123`, then that scene name should take priority over whatever one lies above line number 123.

This should be a pretty quick fix, I believe all you need to do is check whether run_config already has scene_names defined, and if so, don't do the search for the scene.

Varniex:
Thanks @3b1b for the feedback.
I believe the recent changes is what you were suggesting.

manimlib/extract_scene.py
```diff
@@ -142,35 +143,48 @@ def get_indent(code_lines: list[str], line_number: int) -> str:
     return n_spaces * " "
 
 
-def insert_embed_line_to_module(module: Module, line_number: int):
+def insert_embed_line_to_module(module: Module, run_config: Dict) -> None:
```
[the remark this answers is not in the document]
Varniex: I thought about the suggestion.
The reason I am searching for the scene name within this specific function is to access the module's lines to locate the keyword `class` above the specifed line number.

Vegeorca:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

Issue #2288, opened 2025-01-01
Brace doesn't work with unicode-math


xiao-ming-hub:
### Describe the bug

When I use `unicode-math` package in `tex_templates.yml` and I use `Brace` to draw a curly brace, an error has occurred.

**Code**:

```yml
# custom_config.yml
tex:
  template: "basic"
```
```yml
# tex_templates.yml
# Line 8 and 9 are added
basic:
  description: ""
  compiler: latex
  preamble: |-
    \usepackage[english]{babel}
    \usepackage[mathrm=sym]{unicode-math}
    \setmathfont{Noto Sans Math}
    \usepackage{amsmath}
    \usepackage{amssymb}
    \usepackage{xcolor}
```
```py
square = Square().set_fill(BLUE_E, 1)
# Here's where the error happens:
brace = Brace(square, UP)
```

**Wrong display or Error traceback**:

```txt
Traceback (most recent call last):
  ...
  File "/.../manimlib/scene/scene.py", line 155, in run
    self.construct()
    ~~~~~~~~~~~~~~^^
  File "/.../tmp.py", line 5, in construct
    brace = Brace(square, UP)
  File "/.../manimlib/mobject/svg/brace.py", line 50, in __init__
    self.set_initial_width(target_width)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/.../manimlib/mobject/svg/brace.py", line 58, in set_initial_width
    for tip, rect, vect in [(self[0], self[1], RIGHT), (self[5], self[4], LEFT)]:
                                      ~~~~^^^
  File "/.../manimlib/mobject/svg/string_mobject.py", line 551, in __getitem__
    return super().__getitem__(value)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^
  File "/.../manimlib/mobject/mobject.py", line 405, in __getitem__
    return self.split().__getitem__(value)
           ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
IndexError: list index out of range
````

### Additional context

The pack `unicode-math` is a convenient way to set the math font, but it also changes the way curly braces are rendered. LaTeX's original method was to use six characters to adjust the length by stretching, but in this package, the length is spliced according to the desired length, resulting in the subscript of `self` not corresponding.

PR #2289, opened 2025-01-01
Fix Brace. (#2288)

xiao-ming-hub:
## Motivation

I want to use package `unicode-math` to set math fonts, but an the error happened(see issues above).

## Proposed changes

- I use `Text` instead of `Tex` to draw brace.

## Test

Just `UpdatersExample` enough.
```bash
manimgl example_scenes.py UpdatersExample
```


3b1b:
When I try this out, the Brace seems misaligned.

image

I think the preferred way to fix this would be to implement it in a way that doesn't assume the particular underlying structure of six pieces.

PR #2299, opened 2025-01-09
Fix --fps type conversion to int

ShinapriLN:
## Motivation

I'm using a variety of frame rates, and it's inconvenient to change the value in `default_config` each time.

## Proposed changes

Fixed type conversion for the `--fps` argument as an integer to improve convenience.

## Test

**Code**:
Ran the command `manimgl example.py SceneName --fps 240`

**Result**:
The fps value is now passed as an integer without any errors, making it easier to adjust frame rates dynamically.

### DIFF
--- a/manimlib/config.py
+++ b/manimlib/config.py
@@ -174,6 +174,7 @@ def parse_cli():
         parser.add_argument(
             "--fps",
             help="Frame rate, as an integer",
+            type=int,
         )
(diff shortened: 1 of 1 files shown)


Splines:
Would love to see this in main as I can currently not pass any value to `--fps` without running into the mentioned conversion error.

PR #2304, opened 2025-01-16
Add missing dependencies to setup.cfg and requirements.txt

ydm:
## Motivation

There are dependencies not specified in `setup.cfg` and `requirements.txt`.

## Proposed changes
Add them.

## Test
`python -m pip install -e . && run`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 addict
 appdirs
+audioop-lts
 colour
 diskcache
 ipython>=8.18.0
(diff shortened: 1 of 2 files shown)


Splines:
Stumbled upon this as well when setting up Manim builds in a GitHub Actions pipeline. However, this was only a problem with Python 3.13 for us, see [these lines](https://github.com/Manim-Notebook/manim-notebook/blob/03884f31887546999d50592bc76c12a819574930/tests/utils/manimInstaller.ts#L130-L131).

ydm:
I also did hit this problem with Python 3.13, didn't test with other versions.

PR #2308, opened 2025-02-04
Fix module exec

### LINKED ISSUE #2307: Module exec leads to crash
The following scene fails on startup:

```py
from dataclasses import dataclass
from manimlib import *


@dataclass
class Coordinates2D:
    x: float
    y: float


class MyScene(Scene):
    def construct(self):
        print("Hello, world!")

```

Start it via

```bash
manimgl "./path/to/bug-test-file.py" MyScene -e 13
```

and you get the following traceback (Python version: `3.12.8`):

```
  File "/home/dominic/dev/splience-manim-videos/venv/bin/manimgl", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/dominic/dev/manim-fork/manimlib/__main__.py", line 26, in main
    reload_manager.run()
  File "/home/dominic/dev/manim-fork/manimlib/reload_manager.py", line 47, in run
    self.retrieve_scenes_and_run()
  File "/home/dominic/dev/manim-fork/manimlib/reload_manager.py", line 81, in retrieve_scenes_and_run
    scenes = manimlib.extract_scene.main(scene_config, run_config)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dominic/dev/manim-fork/manimlib/extract_scene.py", line 173, in main
    module = get_scene_module(
             ^^^^^^^^^^^^^^^^^
  File "/home/dominic/dev/manim-fork/manimlib/extract_scene.py", line 168, in get_scene_module
    insert_embed_line_to_module(module, embed_line)
  File "/home/dominic/dev/manim-fork/manimlib/extract_scene.py", line 162, in insert_embed_line_to_module
    exec(code_object, module.__dict__)
  File ".home.dominic.dev.splience-manim-videos.manimbug", line 5, in <module>
  File "/usr/lib/python3.12/dataclasses.py", line 1275, in dataclass
    return wrap(cls)
           ^^^^^^^^^
  File "/usr/lib/python3.12/dataclasses.py", line 1265, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/dataclasses.py", line 983, in _process_class
    and _is_type(type, cls, dataclasses, dataclasses.KW_ONLY,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/dataclasses.py", line 749, in _is_type
    ns = sys.modules.get(cls.__module__).__dict__
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute '__dict__'. Did you mean: '__dir__'?
```

The latest commit where it works correctly is df1e0674. Then, in the next commit 1a14a6bd it doesn't work anymore. Thus, #2262 has introduced this bug. It's not by chance that this PR is called "Refactor scene creation".

Splines:
## Motivation

Fixes #2307 

## Proposed changes
- Manually add the module to `sys` before executing it
- I've also renamed `insert_embed_line_to_module` to `insert_embed_line_to_module_and_exec` to better reflect that this method is actually executing the user code. Please read over the docstrings again as I think they are outdated. I've added two lines beforehand to briefly describe what the method is actually doing.

## Test

```py
from dataclasses import dataclass
from manimlib import *


@dataclass
class Coordinates2D:
    x: float
    y: float


class MyScene(Scene):
    def construct(self):
        print("Hello, world!")
```

If you call this via

```
manimgl "./path/to/bug-test-file.py" MyScene -e 13
```

it should show "Hello, world!", instead of crashing like in #2307.

### DIFF
--- a/manimlib/extract_scene.py
+++ b/manimlib/extract_scene.py
@@ -143,8 +143,11 @@ def get_indent(code_lines: list[str], line_number: int) -> str:
     return n_spaces * " "
 
 
-def insert_embed_line_to_module(module: Module, run_config: Dict) -> None:
+def insert_embed_line_to_module_exec(module: Module, run_config: Dict) -> None:
     """
+    Loads the user code, inserts a self.embed() line at the given line_number
+    and executes the code.
+
     This is hacky, but convenient. When user includes the argument "-e", it will try
     to recreate a file that inserts the line `self.embed()` into the end of the scene's
     construct method. If there is an argument passed in, it will insert the line after
@@ -191,7 +194,7 @@ def insert_embed_line_to_module(module: Module, run_config: Dict) -> None:
 def get_module(run_config: Dict) -> Module:
     module = ModuleLoader.get_module(run_config.file_name, run_config.is_reload)
     if run_config.embed_line:
-        insert_embed_line_to_module(module, run_config)
+        insert_embed_line_to_module_exec(module, run_config)
     return module
 
 
--- a/manimlib/extract_scene.py
+++ b/manimlib/extract_scene.py
@@ -174,7 +174,17 @@ def insert_embed_line_to_module(module: Module, run_config: Dict) -> None:
 
     # Execute the code, which presumably redefines the user's
     # scene to include this embed line, within the relevant module.
-    code_object = compile(new_code, module.__name__, 'exec')
+    # Note that we add the user-module to sys.modules to please Python builtins
+    # that rely on cls.__module__ to be not None (which would be the case if
+    # the module was not in sys.modules). See #2307.
+    if module.__name__ in sys.modules:
+        log.error(
+            "Module name is already used by Manim itself, "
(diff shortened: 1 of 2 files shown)


Splines:
Unfortunately, it turned out that this is not a way to go since this breaks the reloading mechanism, i.e. during reload you will run into the newly added error message `Module name is already used by Manim itself, please use a different name`.

3b1b:
Should I close the PR then? Would you like to make a second just with the clarification on naming/docs?

PR #2319, opened 2025-03-02
Create python-package.yml

Jesse-jude:
This will help run the python package on all versions of python

## Proposed changes

- 
- 
- 

## Test

**Code**:

**Result**:


Jesse-jude:
It gives the packages the ability to run this package on all versions of python

PR #2330, opened 2025-04-04
Resolving minor bug in StreamLines

Varniex:
## Proposed changes

- Getting different input parameter in `self.init_style()` from `self.draw_lines()`, by removing the asterisk (*), the parameter `coords` would be passed as a numpy array in both functions.
- Setting a default value to 0 in `AnimatedStreamLines`. As `update()` function is called (with no arguments) in `mobject.py` (L845)


3b1b:
Thanks!

PR #2331, opened 2025-04-05
A necessary tips for using latex added in README

mryuans:
## Motivation
When I first tried to use the LaTeX in my video, I constantly ran into a "LaTeX compilation failed" error. After read the `tex_file_writing.py` in `manimlib/utils`, I figured out that there is a necessary change for temporary file path, which is not obvious for users with their first trial. So I think it's better if it could be add to the README

## Proposed changes
```README.md
#### Tips for using LaTeX

Before using functions that need LaTeX, please change the  
temporary directory for intermediate files (which will lead to "LaTeX compilation failed")
you can change the `tex_file_writing.py` in `manimlib/utils':  
  ```tex_file_writing.py in full_tex_to_svg
  # Write intermediate files to a temporary directory
  with tempfile.TemporaryDirectory() as temp_dir:

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ manim-render example_scenes.py OpeningManimExample
 
 Before using functions that need LaTeX, please change the  
 temporary directory for intermediate files (which will lead to "LaTeX compilation failed")
-you can change the `tex_file_writing.py` in `manimlib/utils':  
+you can change the `tex_file_writing.py` in `manimlib/utils`:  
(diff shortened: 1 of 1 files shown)


3b1b:
Hmm, it sounds like the use of tempfile.TemporaryDirectory() causes an error on your system? Rather than having the Readme suggest people change the source code, we should just fix this error.

Do you have any added details on the nature of the problem you ran into, and what tempfile.TemporaryDirectory() returns for you?

PR #2403, opened 2025-10-30
Fix: remove gradient color artifacts by earcut-based triangulation

### LINKED ISSUE #2387: Feature: Add Workflow for Manual Version Bumps
You should add a GitHub Actions workflow that allows you to **manually trigger version bumps**. This will help you keep your version and changelog consistent in the project.  

**Proposed Workflow**  
```yaml
name: Version Bump

permissions:
  contents: write
  pull-requests: write

on:
  workflow_dispatch:
    inputs:
      release_type:
        description: 'Select version bump type'
        required: true
        default: 'patch'
        type: choice
        options:
          - patch
          - minor
          - major

jobs:
  bump:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v5
        with:
          fetch-depth: 0

      - name: Universal Version Bump
        uses: taj54/universal-version-bump@v0.8.2
        with:
          release_type: ${{ inputs.release_type }}
          git_tag: false
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

### ✅ Acceptance Criteria

- [ ] You can see the workflow in the **Actions** tab.  
- [ ] You can choose bump type (`patch`, `minor`, `major`).  
- [ ] The workflow bumps the version .  
- [ ] (Optional) You can extend this workflow later to include tagging + release creation.

ZaildarAnmol:
Problem: gradient fills show extra colors due to overlapping triangle fan.
Fix: non-overlapping triangulation using mapbox_earcut (outer ring + holes).
Fallback to previous triangulation if dependency absent.
Tested on the minimal scene from issue #2387/#2382 and several concave/holed shapes; artifacts are gone.
No API changes; performance comparable for typical glyphs/shapes.

Fixes #2387

### DIFF
--- a/manimlib/mobject/types/vectorized_mobject.py
+++ b/manimlib/mobject/types/vectorized_mobject.py
@@ -44,6 +44,7 @@ from manimlib.utils.space_ops import rotation_matrix_transpose
 from manimlib.utils.space_ops import poly_line_length
 from manimlib.utils.space_ops import z_to_vector
 from manimlib.shader_wrapper import VShaderWrapper
+from manimlib.utils.triangulation import triangulate_polygon_with_holes
 
 from typing import TYPE_CHECKING
 from typing import Generic, TypeVar, Iterable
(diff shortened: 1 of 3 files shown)


mathsvisualization:
Is he doing the right thing? I don't like it? If it works then also show its results. If I don't want to see this then you can use `set_color_by_code` like this. 
Checkout this #2424

PR #2421, opened 2026-02-09
Added Typst equivalents of TeX and TexText files.

Varniex:
## Motivation

Using [Typst](https://typst.app/) is cleaner, faster, more flexible and less complex than using standard $LaTeX$ (like MikTex) for equations.  Besides the fact that Typst is similar to Markdown, it doesn't need different compilers (and/or packages) for different fonts and other features.

## Proposed changes

- Added a separate file `typst_tex_mobject.py` where the classes are implemented.
- Added typst template in `tex_templates.yml`

## Test

To test this version:
- Install [Typst CLI](https://github.com/typst/typst#installation) for your device.
- Change the default tex template setting to `typst` in `default_config.yml` (or `custom_config.yml`).
- Use `TypstTex` (or `TypstTexText`).

** Some things to apply to keep indexing correct:
1. Use `frac(num, denom)` instead of `num / denom`
2. Always apply superscript (`^`) before subscript (`_`). For eg: `integral^infinity_0 e^(-x^2) d x`

For e.g:
```python
from manimlib import *
from manimlib.mobject.svg.typst_tex_mobject import TypstTex, TypstTexText


class EquationScene(Scene):
    def construct(self):
        equation = TypstTex("E = m c^2")
        self.add(equation)

        text = TypstTexText("Hello, World!")
        self.add(text)
```

Hi @3b1b, if you find this helpful as well, the functions can be moved to `utils/tex_file_writing.py` file.


Varniex:
Awesome!

SalekDev:
@Varniex How would you implement an equation like `F = ma` that has a bold, red and bigger size F? (I already tried defining a custom size and color function which works in typst but it does not in manim, the displayed text in the animation is still just color white).

Varniex:
You can modify the equation just like any Mobject with indexing.
```python
eq = TypstTex("F = m a") # or TypstTex("bold(F) = m a")
eq["F"].set_color(RED).set_stroke(width=1) # or eq[0].set_color(RED)
```

SalekDev:
I didn't know about that way of editing mobjects, thanks!

Varniex:
Opening a new one with the updated ManimGL Commit to ensure no bugs would be in the PR.

PR #2450, opened 2026-03-30
Using Typst instead of LaTeX in `Tex` and `TexText`

Varniex:
## Motivation

Using [Typst](https://typst.app/) is cleaner, faster, more flexible and less complex than using standard $LaTeX$ (like MikTex) for equations.  Besides the fact that Typst is similar to Markdown (with LaTeX like equation format), it doesn't need different compilers (and/or packages) for different fonts and other features.

## Proposed changes

- Changed `tex_mobject.py`, `tex_file_writing.py` and `tex_to_symbol_count.py` to adapt Typst related equation strings
- Changed other related files (like `special_tex.py`) to accept Typst like equations.
- Changed respective code in `example_scenes.py`

## Test

To test this version:
- Install [Typst CLI](https://github.com/typst/typst#installation) for your device.
- Use `Tex` (or `TexText`) as before but with Typst style (like `integral` instead of `\int`).
- To set a different font, you can add these lines in the `templates.yml` or give as the `additional_preamble` parameter:
    1. TexText Font: `#set text(font: <font_name>)`
    2. Tex Font: `#show math.equation: set text(font: <math_font_name>)`

- Some things to apply to keep indexing correct:
    1. Use `frac(num, denom)` instead of `num / denom`
    2. Always apply superscript (`^`) before subscript (`_`). For eg: `integral^infinity_0 e^(-x^2) d x`

For e.g:
```python
from manimlib import *


class EquationScene(Scene):
    def construct(self):
        equation = Tex("E = m c^2")
        self.add(equation)

        text = TexText("Hello, World!")
        self.add(text)
```


jeremyong:
FYI it looks like the opening scene in this PR doesn't compile since `rightarrow` in the most recent Typst is spelled `arrow` or just `->` but otherwise, this is working for me. Cool patch!

PR #2465, opened 2026-05-13
Fix/code class pygments short hex #2355

### LINKED ISSUE #2355: Code Class not work
### Describe the bug
the Code Class not work. just raise Exception.

class DemoScene(Scene):

    def construct(self):
        
        c = Code('sorted(iterable, key=None, reverse=False)')
        self.play(FadeIn(c))
        
        self.wait()

**Wrong display or Error traceback**:

leonor-p:
## Motivation
The `Code` class raises an exception when instantiated with any string.
`PangoMarkupFormatter` in pygments >= 2.14 generates 3-digit hex color
codes (e.g. `#abc`) in markup attributes. Pango does not support short
hex notation, causing `MarkupText` to fail on instantiation.

Fixes #2355

## Proposed changes
- Added regex substitution in `Code.__init__` in `manimlib/mobject/svg/text_mobject.py` that expands all 3-digit hex colors to 6-digit after `pygments.highlight()`, before passing markup to `MarkupText`
- Added unit tests in `tests/test_code_class.py` verifying the regex expansion works correctly on Python 3.9

## Test
**Code**:
```python
c = Code('sorted(iterable, key=None, reverse=False)')
```

**Result**: Code object created successfully without exception.
Unit tests pass on Python 3.9.25 — 2 passed in 0.02s.
image


3b1b:
Looks reasonable, but do you mind updating the PR to remove the stray files, and implement it by defining the expand_hex function, and calling it on that one tex_mobject.py line insetad of putting the regex logic in place.

PR #2482, opened 2026-08-02
Cleaner rending logic

3b1b:
Fill was computing winding numbers by abusing alpha blending, which was fragile and wrong for translucent self-overlap. It now counts them in a stencil buffer, giving correct blending and real per-fragment depth.

Geometry shaders are gone. Dots, fills and strokes all expand from a fixed vertex budget in the vertex shader, reading records from the vertex buffer as a texture. They cost 38.7 µs per draw against 0.68 on macOS, where they're emulated.

Batching is gone. Every mobject gets its own buffer and its own draw calls, which is what makes interleaving correct and removes render groups entirely. Paying for that needed uniforms sent as one block per mobject, shader source cached against its layout, and clip planes left permanently enabled.

Stroke joints are computed in the shader from the subpath a point belongs to, rather than stored per point. Joints that would jut out are rounded off using vertex budget the polyline never used, so a grazing-angle polygon no longer throws spikes. joint_type's four options — two of them misnamed, one broken — become a single joint_roundness from 0 to 1.

Fill colour is a uniform, and a second one gives gradients, defined by the two points they run between. Because those are points, a gradient shifts, scales and rotates with its mobject, and a gradient set on a word runs across the whole of it.

Simplified Surface. No more carrying unit normals around, and now there's a better solution for avoiding glitches during transparency than having to use always_sort_to_camera.

StructuredArray: Each mobject's data and uniforms are now both held in a small wrapper around a structured numpy array, laid out to mirror what its shaders read, which records changes to its data and remembers state when resized to 0. This replaces the need for Mobject._data_has_changed, Mobject._data_defaults, the @affects_data and @affects_family_data decorators.

Other Removals: 
- The ear-clipping triangulation that the stencil count replaces
- Mapbox-earcut
- The shader wrapper's id, left over from batching
- Duplicate record of where subpaths lie.
- refresh_shader_wrapper / @affects_shader_wrapper
- Use of render primitives (TRIANGLE_STRIP, etc.)


MathItYT:
Hi! I was testing your new branch and there's a little problem when a VMobject has opacity less than one.

```py
from manimlib import *

class MyScene(Scene):
    def construct(self):
        # Create a simple text object
        text = Tex(r"\mathbb{R}").set_fill(opacity=0.5)
        text.set_height(4)
        self.add(text)
```

MyScene

It should blend the R's opacity over background, but the final opacity is replaced by R's opacity instead.

The same with stroke and anti-aliasing stuff. I hope it's useful.

3b1b:
Thanks for the comment; that should be an easy fix.

MathItYT:
Perfect! As we were talking about VMobject's rendering, could I ask you to allow both fill and stroke have their own `set_fill_by_code`/`set_stroke_by_code`, if possible? That would allow me to code my complex function color plotting video and even better to add a texture (like image clipping with VMobject's rendered shape).

3b1b:
Does set_color_by_code not work for you? It doesn't split out for fill and stroke; it would affect both equally, but presumably that's what you'd want.

Edit: I'm realizing, looking at it now, that it sets the color per vertex, not per fragment. Making it per fragment is more clearly what one would want.

MathItYT:
Nope, look at this:

```py
import manimlib as m


class MyInteractiveScene(m.InteractiveScene):
    def setup(self) -> None:
        super().setup()
        self.camera.background_rgba = [0, 0, 0, 0]

    def construct(self) -> None:
        txt = m.Circle().set_fill(m.WHITE, opacity=1)
        txt.set_color_by_code("""frag_color.a *= 0.5;""") # glsl
        self.add(txt)
```

```
ManimGL v1.7.2
  File "/home/mathlike/Escritorio/bot/.venv/bin/manimgl", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/__main__.py", line 61, in main
    run_scenes()
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/__main__.py", line 39, in run_scenes
    scene.run()
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/scene/scene.py", line 162, in run
    self.tear_down()
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/scene/scene.py", line 179, in tear_down
    self.file_writer.finish()
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/scene/scene_file_writer.py", line 197, in finish
    self.scene.update_frame(force_draw=True)
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/scene/scene.py", line 249, in update_frame
    self.camera.capture(*self.mobjects)
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/camera/camera.py", line 285, in capture
    mobject.render(self.ctx)
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/mobject/mobject.py", line 2114, in render
    shader_wrapper = mob.get_shader_wrapper(ctx)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/mobject/mobject.py", line 2093, in get_shader_wrapper
    self.init_shader_wrapper(ctx)
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/mobject/types/vectorized_mobject.py", line 1261, in init_shader_wrapper
    self.shader_wrapper = VShaderWrapper(
                          ^^^^^^^^^^^^^^^
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/shader_wrapper.py", line 330, in __init__
    self.replace_code_program(old, new, program_type)
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/shader_wrapper.py", line 395, in replace_code_program
    super().replace_code(old, new)
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/shader_wrapper.py", line 168, in replace_code
    self.init_program()
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/shader_wrapper.py", line 344, in init_program
    self.stroke_program = get_shader_program(
                          ^^^^^^^^^^^^^^^^^^^
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/manimlib/utils/shaders.py", line 79, in get_shader_program
    program = ctx.program(
              ^^^^^^^^^^^^
  File "/home/mathlike/Escritorio/bot/.venv/lib/python3.12/site-packages/moderngl/__init__.py", line 1978, in program
    res.mglo, _members, res._subroutines, res._geom, res._glo = self.mglo.program(
                                                                ^^^^^^^^^^^^^^^^^^
_moderngl.Error: GLSL Compiler failed

vertex_shader
=============
0(200) : error C1503: undefined variable "frag_color"
```

3b1b:
Don't make it a function of frag_color, instead just edit the variable "color".  Take a look at finalize_color.glsl in the shaders folder; the way this all works is by inserting your code into that function.

MathItYT:
Oh, my bad! Thanks, now it works fine, but I wanted to put a black stroke around my color plot. If you could separate fill and stroke logic would be great, please :)

3b1b:
Why not just have a separate mobject you render just with a black stroke?

MathItYT:
Oh, nice solution! Although if we talk about performance, it will be a tiny bit slower than separating logic for one mobject

3b1b:
I doubt it would affect performance much if at all. When a VMobject has both stroke and fill, those render calls are separate, so it's about the same as if you have one with just fill and add another with just stroke.

MathItYT:
You're right. Anyways, thanks for your patience! VMobject's rendering improved significantly and it feels more accurate (I mean output seems correct, without those artifacts) :)

PR #2484, opened 2026-08-04
Prepare the renderer for the wgpu port

3b1b:
- **Count writes to a `StructuredArray`** rather than saying whether it changed. A count says what a watcher has missed instead of whether anyone has, so more than one thing can watch one array, and nothing has to put a flag back. 
- Rounds unit normals, so mobjects facing the same way say so exactly.
- **A frame comparison harness** (`tests/render_compare.py`). There is no way to unit test a renderer; what there is, is the frame it drew last time. 
- **manim's own names for keys and modifiers** (`event_keys.py`). Three files above the window imported pyglet's constants; now the window translates and nothing above it knows what it is built on. Collapses a ten-times-repeated `MOD_COMMAND | MOD_CTRL` into `Mods.CTRL_OR_CMD`.
- **A `Renderer` handed to a mobject** in place of the bare moderngl context, which is where what holds for a whole frame can live.
- **The camera's uniforms as one std140 block**, exactly as a mobject's have travelled since #2482. Deletes `ALL_PROGRAMS`, `SHARED_UNIFORMS`, `set_shared_uniforms`, `get_shared_uniform` and the loop pushing every shared value into each newly compiled program. All of these existed because GL has no way to say a value holds for everything. 
- **Clipping by dropping fragments** rather than by writing `gl_ClipDistance`. wgpu has no clip distance on the backends this is headed for.
- **Named draw states, and every write before any draw.** A fill's forty lines of `glStencilOpSeparate`/`glColorMask` poking become three states asked for by name; there are six states in all, which is the useful thing to have learned. And `pre_render` splits into `write_buffers` and `render`, because a single wgpu render pass per frame requires every buffer write to precede it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -156,3 +156,6 @@ CLAUDE.md
 
 # LaTeX scratch files, regenerated whenever Tex is rendered
 latex_cache/
+
+# Textures the render comparison harness makes for itself
+tests/assets/
--- a/manimlib/camera/camera.py
+++ b/manimlib/camera/camera.py
@@ -276,8 +276,19 @@ class Camera(object):
         # only here to guarantee a clean slate, e.g. if a previous frame's
         # rendering was interrupted partway through
         gl.glClear(gl.GL_STENCIL_BUFFER_BIT)
-        for mobject in mobjects:
-            mobject.render(self.renderer)
+
+        wrappers = [
+            wrapper
+            for mobject in mobjects
+            for wrapper in mobject.get_shader_wrappers(self.renderer)
+        ]
+        # Everything the frame sends to the gpu, before anything it draws, see
+        # ShaderWrapper.write_buffers
+        for wrapper in wrappers:
+            wrapper.write_buffers()
+        self.renderer.reset_state()
+        for wrapper in wrappers:
+            wrapper.render()
 
         if self.window:
             self.window.swap_buffers()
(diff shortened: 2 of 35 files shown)


MyDum-bsu:
Hey Grant! Amazing how the library is improving so rapidly.

Small suggestion for `VShaderWrapper`:

`Mobject.shader_folder` is honoured by every mobject kind (`Surface`, `ImageMobject`, base `ShaderWrapper.get_code`), but `VShaderWrapper.init_program_code` hardcodes `"quadratic_bezier"` and `VMobject.init_shader_wrapper` never passes `shader_folder` through. So custom stroke/fill shaders for VMobjects are impossible, while the plumbing exists everywhere else.

Fix is two lines, behaviour-preserving:

```python
# VShaderWrapper.init_program_code
base_folder = self.shader_folder or "quadratic_bezier"
# ... os.path.join(base_folder, f"{vtype}", f"{name}.glsl")
```

```python
# VMobject.init_shader_wrapper
shader_folder=self.shader_folder,  # pass to VShaderWrapper(...)
```

Default stays `quadratic_bezier` (empty string is falsy).

I'm building glow objects like `GlowText`, `GlowTex`, `GlowVMobject`, `GlowGroup` as VMobject subclasses. Part of that is a glow rendered directly in the stroke shader, which needs a custom `stroke`/`fill` shader folder per VMobject subclass. `shader_folder` is the only hook for that, and it's currently dead for VMobjects.

MyDum-bsu:
One more small one:

`SceneFileWriter` takes `saturation` and `gamma`, and applies them to video output via `eq` in `open_movie_pipe` (`vf_arg += f',eq=saturation={self.saturation}:gamma={self.gamma}'`).
But `save_final_image` (`-s`, save-last-frame PNGs) ignores them. The saved frame comes out with the raw values.

Issue #2471, opened 2026-06-18
Embed appears to cause manim to peg an entire CPU core


scuff3drook:
### Describe the bug
From what I can tell, compared to the typical event loop during scene playback, after the `embed` event loop kicks in, `manim` will start to completely peg a core, resulting in elevated CPU usage, battery usage (on laptop) or thermal throttling.

### Additional context
I only did a brief investigation but am not yet sure what might be causing the event loop sleep timer or vsync to get disrupted by the embed. I decided to file the issue first and investigate more later -- pointers appreciated.

hiSandog:
A minimal scene and the exact Manim/Python/backend versions would make this easier to reproduce. A short `py-spy` or cProfile capture during the embedded loop could confirm whether it is busy-polling input, repeatedly rendering, or missing a sleep/vsync wait. Comparing interactive and non-interactive renderers with the same scene would narrow the event-loop boundary further.

PR #2485, opened 2026-08-05
Fixed pegging CPU Core Issue

ithun-y-ittesaf:
## Motivation

As described in #2471 

After entering the interactive embed (`self.embed()`), manim's IPython input hook busy-loops while waiting for terminal input, with no sleep in the loop. This pegs a full CPU core for as long as the embedded prompt sits idle, causing elevated CPU/battery usage and, on laptops, thermal throttling. 

This differs from normal scene playback, where `Scene.interact()` drives
`update_frame` with a nonzero `dt`, which sleeps internally to line up with the target frame rate ("vsync"). The embed's input hook always calls `update_frame(dt=0)`, which skips that sleep and returns immediately when there's no pending window event, so the `while` loop spins as fast as the interpreter allows.

## Proposed changes
- In `InteractiveSceneEmbed.enable_gui`'s `inputhook`, sleep between polls instead of spinning unbounded, following the same pattern IPython's own bundled `pt_inputhooks/pyglet.py` uses for pyglet-based GUIs (which explicitly calls out this exact 100%-CPU failure mode).
- Back off the poll interval the longer the shell has been idle (1ms → 50ms → 1s) so quick window interactions (e.g. dragging/rotating the scene from the embed) still feel immediate, while genuinely idle time doesn't burn a core.
- When the window is closing, `break` out of the loop immediately instead of continuing to spin until input becomes ready.

## Test
**Code**:
```python
class Test(Scene):
    def construct(self):
        self.embed()
```

Run with `manimgl` and let the embedded IPython prompt sit idle (no typing, no window interaction) for ~15-20 seconds, watching CPU usage for the process (e.g. top/Activity Monitor).

Result:

Before: one core pegged at ~100% the entire time the embed is idle.
After: CPU usage for the process drops close to idle (brief high-frequency polling immediately after a cell runs, backing off within ~100ms, down to a check once per second after ~10s of inactivity). Window responsiveness (mouse drag/scroll/rotate) during the embed is unaffected.

### DIFF
--- a/manimlib/scene/scene_embed.py
+++ b/manimlib/scene/scene_embed.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import inspect
 import pyperclip
 import textwrap
+import time
 import traceback
 
 from IPython.terminal import pt_inputhooks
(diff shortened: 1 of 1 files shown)


3b1b:
Thanks! I agree with the issue, but instead of the progressive addition to the sleep time, we can just lock it based on the intended frame rate for the scene.

PR #2483, opened 2026-08-04
Smooth out two sources of hard edges in VMobject rendering

MathItYT:
## Motivation
VMobject's rendering has significantly improved its rendering, but there was still something to improve: anti-aliasing.

I wrote the following Python script:

```py
import manimlib as m


class MyInteractiveScene(m.InteractiveScene):
    drag_to_pan: bool = False

    def setup(self) -> None:
        super().setup()
        self.camera.background_rgba = [0, 0, 0, 1]

    def construct(self) -> None:
        linear_algebra = m.TexText("Álgebra Lineal", additional_preamble=r"\usepackage{mlmodern}").scale(3).to_edge(m.UP).shift(m.LEFT * 2).rotate(7*m.DEGREES)
        matrix_times_vector = m.Tex(r"\mathbf{A}\mathbf{x} = \mathbf{b}", additional_preamble=r"\usepackage{mlmodern}").scale(3).to_edge(m.DOWN, buff=1).shift(3*m.RIGHT)
        matrix_times_vector.rotate(14*m.DEGREES)
        num_plane = m.NumberPlane(x_range=[-5, 5, 1], y_range=[-2, 2, 1])
        p1_1 = [-5, -1, 0]
        p1_2 = [5, 1.5, 0]
        p2_1 = [-5, 1.5, 0]
        p2_2 = [5, -0.5, 0]
        coefficients_1 = [
            (p1_2[1] - p1_1[1]) / (p1_2[0] - p1_1[0]),
            p1_1[1] - ((p1_2[1] - p1_1[1]) / (p1_2[0] - p1_1[0])) * p1_1[0],
        ]
        coefficients_2 = [
            (p2_2[1] - p2_1[1]) / (p2_2[0] - p2_1[0]),
            p2_1[1] - ((p2_2[1] - p2_1[1]) / (p2_2[0] - p2_1[0])) * p2_1[0]
        ]
        intersection_x = (coefficients_2[1] - coefficients_1[1]) / (coefficients_1[0] - coefficients_2[0])
        intersection_y = coefficients_1[0] * intersection_x + coefficients_1[1]
        line1 = m.Line(start=p1_1, end=p1_2, stroke_color=m.YELLOW, stroke_width=4)
        line2 = m.Line(start=p2_1, end=p2_2, stroke_color=m.YELLOW, stroke_width=4)
        dot = m.GlowDot(center=[intersection_x, intersection_y, 0], radius=0.5, glow_factor=2.0, color=m.YELLOW)
[...]

PR #2486, opened 2026-08-05
Port the renderer from OpenGL to wgpu

3b1b:
ManimGL's rendering now goes through wgpu rather than ModernGL, and its windowing through rendercanvas/glfw rather than pyglet. Shaders are WGSL, one file per kind of mobject rather than a folder of stages.

The core motivation is that Apple stopped supporting OpenGL, so a port of the backend rendering like this has been long overdue. In theory, it should not affect users, except insofar as there is now a different set of requirements.

- Out: `moderngl`, `moderngl_window`, `PyOpenGL` and `screeninfo`
- In: `wgpu`, `rendercanvas` and `glfw`

For anyone engaged with writing custom shaders, it does have a pretty meaningful change, which is the update from those shaders being written in .glsl to .wgsl.  In Mobject, `shader_folder` becomes `shader_file` and names a single file.

## What a frame does now

A frame is one render pass. Everything it sends to the gpu happens before that pass opens, because a write reaching the gpu partway through a pass has no say over which draws see it.

Mobject data and uniforms are now gathered together in a shared "Area" objects to reduce the number of buffer writes. Two arenas per record size (one for uniforms, one for records) hand out a stretch to each mobject for the length of a frame and go out in a single write.

## That the picture is the same

`tests/render_compare.py` renders 29 scenes and compares them against references captured from master. Five are byte-identical. Of the 24 that differ, 22 differ only on the edges of shapes, where two rasterizers never agree; the other two differ by at most 4/255 across a few hundred interior pixels. Rolling a render by one pixel makes every difference 8-70x worse, so nothing is offset, and the signs are balanced.

Two more tests need no reference picture: `tests/readback.py` checks that every frame drawn reaches the file exactly once and unaltered, and `tests/surface_order.py` checks that a see-through surface looks the same wound two different ways.

Fixed along the way: a fill's border now closes around its own subpaths, so an open path's closing edge is anti-aliased rather than hard (reported by @MathItYT, #2483).


MathItYT:
Hey, Grant! Nice change! I'm now developing a branch to port your Manim to the web through the rendercanvas Pyodide backend and replacing subprocesses with MathJax, CanvasKit WASM, etc., are you interested on merging my pull request when I'm done?

3b1b:
I'm hesitant to make any promises, but I do think some of those changes would be in the right direction, e.g. usage of MathJax.

Vipitis:
Hey Grant,

wgpu-py contributor here. First of all awesome to see adoption in such a major project!
As long as you run on native platforms, glsl code is still supported (although it gets translated to wgsl and then again to the selected backend). I hope to get OpenGL passthrough in the next update, so if you run on the OpenGL backend with GLSL code, you can keep that for debugging in renderdoc etc. 

I have plans and a working prototype to make wgpu-py run in the browser with pyodide (basically calling webgpu via js ffi instead of wgpu-native). The design goal is 100% portable code between the native python and the browser run python. But of course there is other pyodide limitations... Maybe I find time to hack in preview wheels and give it a try.

3b1b:
Hi Vipitis, that's good to know. I think with this port, there is no longer any need to run an OpenGL backend or GLSL. I feel like if it's worth porting, it's worth going all the way to keep as clean a final result as possible.

Also, love the idea of having things run in the browser!

PR #2490, opened 2026-08-10
Honor stroke widths declared in SVG files

### LINKED ISSUE #2343: SVGMobject sets stroke widths to 0.0 by default
### Description

If an SVG has a path with a stroke, `SVGMObject` will fail to render that stroke because by default, it sets the stroke width of all paths to 0.0

### Repro code

This is a simple SVG that can illustrate the problem:

```svg
<?xml version="1.0" encoding="UTF-8"?>
<svg id="a" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" width="100" height="100" viewBox="0 0 100 100">
  <g id="b">
    <path
      d="M16.13,2.91c17.76-1.74,44.96-.65,63.35-.15,1.91.05,6.98-.25,8.07,1.24,1.1,30.24.07,60.73.49,91.06-.23,1.42-.31,2.29-1.88,2.82l-71.51.15c-1.48-.19-2.6-.9-2.72-2.47V5.49l1.3-2.42h0c.26-.13,2.32-.09,2.9-.15Z"
      style="fill: #fac328; stroke: #404140; stroke-linecap: round; stroke-linejoin: round; stroke-width: 1.75px;"/>
  </g>
</svg>
```

I've saved that in a file called `Simple.svg`, which I load into scene with this code:

```py
from manimlib import *

class MatrixTransform(Scene):
    def construct(self):
        p = SVGMobject("Simple.svg")
        self.add(p)
```

I run that with this command:

`manimgl .\test_svg.py`

It reports that I'm running `v1.7.2`)

### Wrong display**

I see this output:

Image

(Note, I'm using `camera.background_color: "#88BBCC"` in my custom settings, hence the non-standard background colour.)

Here's how the SVG should be rendered:

Image

### Workaround

It is possible to work around this by explicitly setting the `stroke_width` to `None` when creating the `SVGMobject`:

```py
from manimlib import *

class MatrixTransform(Scene):
    def construct(self):
        p = SVGMobject("Simple.svg", stroke_width=None)
        self.add(p)
```

With that setting, I get this:

Image

### Additional context

The problem appears to be this line of code here:

https://github.com/3b1b/manim/blob/f4737828f65999252c0cab9824d9d34cfabf4341/manimlib/mobject/svg/svg_mobject.py#L54

If `stroke_width` is set to the default of `0.0`, `SVGMobject` seems to interpret this as an instruction to set the stroke width to `0.0` on every path in the SVG, regardless of what the SVG might say. If the default here were `None` (like other arguments such as `stroke_color` and `stroke_opacity`) then by default it would render the SVG without modification, but scenes that want to override the `stroke_width` can still do so

00PrabalK00:
## Motivation

Any SVG with a stroked path draws with no stroke at all. `SVGMobject` defaults `stroke_width` to `0.0` and applies it to the whole family after parsing, so the per path widths read from the file get wiped out.

The default was doing real work, which is why it cannot simply go: svgelements reports a stroke width of `1.0` even for shapes painted with `stroke: none`, so unstroked shapes would pick up a spurious outline. Widths also need converting, since a file states them in its own user units while the geometry gets resized to the requested height or width, and a `stroke_width` of 100 spans one unit of a default scale frame (`STROKE_WIDTH_CONVERSION` in `shaders/stroke.wgsl`).

Fixes #2343

## Proposed changes

- Take a shape's stroke width only when a stroke color came with it, and zero it otherwise, so the default can become `None` and parsed strokes survive
- Convert those widths for the units the file is drawn in and the size the geometry is given
- Apply the caller's style after the resizing, so an explicit `stroke_width` is the width drawn. `Tex`, `Text` and the bundled SVG drawings are unaffected

## Test

**Code**:

The SVG from #2343, on a light background so the dark stroke reads:

```py
from manimlib import *

class IssueRepro(Scene):
    def construct(self):
        bg = Rectangle(width=FRAME_WIDTH, height=FRAME_HEIGHT)
        bg.set_fill("#88BBCC", 1).set_stroke(width=0)
        self.add(bg, SVGMobject("issue.svg", height=5))
```

**Result**:

Before, `get_stroke_width()` on the path is `0.0` and nothing draws:

before

After, the width is `9.05` and the stroke draws at the thickness the file asks for:

after

For the units, two files drawing the same picture, one in a 100 unit viewBox with `stroke-width: 4`, the other in a 1000 unit viewBox with `stroke-width: 40`, both at `height=3`. They now report the same width of `13.04` and draw alike, where without the conversion they came out `4.0` and `40.0`:

scale

The other side of the change: `Text("Regression check")` still reports stroke width `0.0` on every glyph and renders identically, and a few multi path SVGs from matplotlib come out with widths only on the paths that declare a stroke.

### DIFF
--- a/manimlib/mobject/svg/svg_mobject.py
+++ b/manimlib/mobject/svg/svg_mobject.py
@@ -27,6 +27,10 @@ if TYPE_CHECKING:
     from manimlib.typing import ManimColor, Vect3Array
 
 
+# A stroke_width of 100 spans one unit of a default scale frame, see
+# STROKE_WIDTH_CONVERSION in shaders/stroke.wgsl
+STROKE_WIDTHS_PER_UNIT: float = 100.0
+
 SVG_HASH_TO_MOB_MAP: dict[int, list[VMobject]] = {}
 PATH_TO_POINTS: dict[str, Vect3Array] = {}
 
@@ -66,7 +70,7 @@ class SVGMobject(VMobject):
         color: ManimColor = None,
         fill_color: ManimColor = None,
         fill_opacity: float | None = None,
-        stroke_width: float | None = 0.0,
+        stroke_width: float | None = None,
         stroke_color: ManimColor = None,
         stroke_opacity: float | None = None,
         # Style that fills only when not specified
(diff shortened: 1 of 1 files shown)


3b1b:
Looks good, thanks!

PR #2492, opened 2026-08-10
Clean Animation.begin

3b1b:
- Prepare the ends before the first blend, not after. 
- Skip a bounding box interpolation when start/end match

### DIFF
--- a/manimlib/animation/animation.py
+++ b/manimlib/animation/animation.py
@@ -74,10 +74,26 @@ class Animation(object):
             self.mobject_was_updating = not self.mobject.updating_suspended
             self.mobject.suspend_updating()
         self.families = list(self.get_all_families_zipped())
+        self.prepare_interpolation()
         self.interpolate(0)
+
+    def prepare_interpolation(self) -> None:
+        """
+        Whatever holds of the two ends for the whole of the animation, settled here rather
+        than found again in every blend, see Mobject.prepare_interpolation.
+
+        The ends are whichever mobjects interpolate_submobject is handed alongside each one,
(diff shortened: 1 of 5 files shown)


00PrabalK00:
Ran the branch against master on a few animations, sampling points and bounding boxes at alpha 0, 0.3, 0.7 and 1: `Transform` with aligned ends, `Transform` with unaligned ends, `ReplacementTransform`, `FadeTransform` and `Write` all come out identical, so the reordering and the box skip look behavior preserving.

One thing worth a look. The new `Animation.prepare_interpolation` decides it has two ends from `len(first) == 3`, and `DrawBorderThenFill` also zips three, its third being the outline rather than an end. So `Write` now goes through `mobject.prepare_interpolation(starting_mobject, outline)`, which it never did before, and its submobjects come out of `begin` with `skip_box_interpolation` set to True.

It happens to be harmless today. The outline is a style only copy of the mobject, so the boxes really do match, phase one starts with `submob.set_data(outline.data)` so the skipped fields already hold what a blend would write, and phase zero goes through `pointwise_become_partial` which refreshes the box from points anyway. Identical frames either way, as above.

Still, it is inferred rather than declared. Any animation that zips three mobjects without the last two being ends gets the treatment silently, and the failure mode would be a stale box or an unwritten field rather than an error. Might be worth an explicit hook, along the lines of the animation naming its ends or overriding `prepare_interpolation` to do nothing, so `DrawBorderThenFill` opts out on purpose instead of by accident.

3b1b:
Yes, good point, and I agree. Addressed here: https://github.com/3b1b/manim/pull/2494

PR #2496, opened 2026-08-11
Keep a moved bounding box's corners the right way round

korjavin:
## Motivation

`apply_points_function` has two ways of keeping a bounding box up to date. Most transforms leave it to be worked out again from the points; `shift`, `scale` and `stretch` pass `works_on_bounding_box=True` and move the two cached corners instead, there being no need to look at every point to know where a box went.

That shortcut assumes the corner which was smallest stays smallest. A negative factor breaks the assumption — it sends the low corner high and the high corner low — and the box is left inside out:

```python
Square().stretch(-1, 0).get_bounding_box()
# [[ 1. -1.  0.]   <- mins
#  [ 0.  0.  0.]
#  [-1.  1.  0.]]  <- maxs
```

So `get_width()` comes out negative, and `get_left()` is to the right of `get_right()`. Everything downstream of the box is then wrong for that mobject: `next_to`, `align_to`, `surround`, `are_points_touching`.

`scale` can't reach this, clipping its factor to `min_scale_factor`, and `flip` goes through `rotate`, which works the box out afresh. `stretch` is the way in, and mirroring with `stretch(-1, dim)` is the ordinary reason to call it that way. `Dodecahedron` already does, at `three_dimensions.py:374`, and looks right anyway only because a group derives its own box from its children's corners with `min` and `max`, which puts an inside out one back the right way round. Ask the stretched face itself and it answers wrongly.

## Proposed changes

Re-derive which corner is which after moving them, rather than assuming the order survived, and put the midpoint between the two:

```python
corners = moved(box[::2].copy())
box[0], box[2] = corners.min(0), corners.max(0)
box[1] = (box[0] + box[2]) / 2
```

Same work for a positive factor, since `min` and `max` of an order which held leave it alone.

## Test

`tests/bounding_box.py`, alongside the other scripts in `tests/`, checking a box against the points it is meant to cover. It runs the two ways in — `stretch` with a negative factor, about the mobject's own center and about another point — along with `shift`, `scale`, positive `stretch` and `rotate`, so the paths which were already right are held to the same thing.

```
python tests/bounding_box.py
```

Before:

```
  stretch(-1, 0): mins [1.0, -1.0, 0.0] for [-1.0, -1.0, 0.0], maxs [-1.0, 1.0, 0.0] for [1.0, 1.0, 0.0]
  stretch(-2, 1) off center: mins [0.0, 2.0, 0.0] for [0.0, -2.0, 0.0], maxs [2.0, -2.0, 0.0] for [2.0, 2.0, 0.0]
  shift: box covers its points
  scale: box covers its points
  stretch(2, 1): box covers its points
  rotate: box covers its points
  FAIL: stretch(-1, 0) left the box saying mins [1.0, -1.0, 0.0] for [-1.0, -1.0, 0.0], maxs [-1.0, 1.0, 0.0] for [1.0, 1.0, 0.0]
  FAIL: stretch(-2, 1) off center left the box saying mins [0.0, 2.0, 0.0] for [0.0, -2.0, 0.0], maxs [2.0, -2.0, 0.0] for [2.0, 2.0, 0.0]
```

After:

```
  stretch(-1, 0): box covers its points
  stretch(-2, 1) off center: box covers its points
  shift: box covers its points
  scale: box covers its points
  stretch(2, 1): box covers its points
  rotate: box covers its points
  a moved box keeps its corners the right way round
```

### DIFF
--- a/manimlib/mobject/mobject.py
+++ b/manimlib/mobject/mobject.py
@@ -272,7 +272,10 @@ class Mobject(object):
             for key in mob.pointlike_uniform_keys:
                 mob.uniforms.apply(key, moved)
             if box is not None:
-                box[:] = moved(box)
+                # Re-derive the corners, since a negative factor swaps which is which
+                corners = moved(box[::2].copy())
+                box[0], box[2] = corners.min(0), corners.max(0)
+                box[1] = (box[0] + box[2]) / 2
 
         if not works_on_bounding_box:
             self.refresh_bounding_box(recurse_down=True)
--- a/tests/bounding_box.py
+++ b/tests/bounding_box.py
@@ -1,0 +1,68 @@
+"""
+That a cached bounding box still says which corner is which after a transform.
+
+Most transforms leave the box to be worked out again from the points, but shift, scale and
+stretch move the cached corners instead, there being no need to look at every point to know
+where a box went. That shortcut assumes the corner which was smallest stays smallest, which a
+negative factor does not: stretch(-1, dim) sends the low corner high and the high corner low,
+and the box comes out inside out, get_width() of it negative and get_left() to the right of
+get_right().
+
+Nothing in the harness reads a box that closely. Dodecahedron builds a face with
+stretch(-1, 2, about_point=ORIGIN) and looks right anyway, a group working its own box out
+from the corners of its children with min and max, which puts an inside out one back the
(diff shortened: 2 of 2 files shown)


3b1b:
Looks good, thanks!

PR #2499, opened 2026-08-12
docs: sync CLI flag reference with current parser

AmeerJ97:
The CLI reference had drifted from the current parser behavior, so this updates the existing table to reflect the supported options.

This change:

- removes the stale `--save_pngs` and `--config` entries;
- adds `--vcodec`, `--pix_fmt`, `--subdivide`, `--show_animation_progress`, `--prerun`, and `--clear-cache`;
- updates the argument forms and descriptions for `--start_at_animation_number` and `--embed`;
- aligns the `--skip_animations` wording with the parser; and
- documents what `--quiet` suppresses.

`START[,END]` is intentionally used as a user-facing description of the accepted value shape rather than argparse's generated `START_AT_ANIMATION_NUMBER` variable name.

Validation:

- every documented long option matches an `add_argument` definition in `manimlib/config.py`;
- docutils parses the complete grid table with no warnings;
- all three table borders have identical widths and no row exceeds the table width; and
- `git diff --check` passes.

A full Sphinx build was not run because the documentation dependencies in this checkout are incomplete.

### DIFF
--- a/docs/source/getting_started/configuration.rst
+++ b/docs/source/getting_started/configuration.rst
@@ -38,7 +38,7 @@ flag                                                       abbr   function
 ``--help``                                                 ``-h`` Show the help message and exit
 ``--version``                                              ``-v`` Display the version of manimgl
 ``--write_file``                                           ``-w`` Render the scene as a movie file
-``--skip_animations``                                      ``-s`` Save the last frame
+``--skip_animations``                                      ``-s`` Skip to the last frame
 ``--low_quality``                                          ``-l`` Render at a low quality (for faster rendering)
 ``--medium_quality``                                       ``-m`` Render at a medium quality
 ``--hd``                                                          Render at a 1080p quality
--- a/docs/source/getting_started/configuration.rst
+++ b/docs/source/getting_started/configuration.rst
@@ -38,31 +38,35 @@ flag                                                       abbr   function
 ``--help``                                                 ``-h`` Show the help message and exit
 ``--version``                                              ``-v`` Display the version of manimgl
 ``--write_file``                                           ``-w`` Render the scene as a movie file
-``--skip_animations``                                      ``-s`` Skip to the last frame
+``--skip_animations``                                      ``-s`` Save the last frame
 ``--low_quality``                                          ``-l`` Render at a low quality (for faster rendering)
 ``--medium_quality``                                       ``-m`` Render at a medium quality
 ``--hd``                                                          Render at a 1080p quality
 ``--uhd``                                                         Render at a 4k quality
 ``--full_screen``                                          ``-f`` Show window in full screen
 ``--presenter_mode``                                       ``-p`` Scene will stay paused during wait calls until space bar or right arrow is hit, like a slide show
-``--save_pngs``                                            ``-g`` Save each frame as a png
 ``--gif``                                                  ``-i`` Save the video as gif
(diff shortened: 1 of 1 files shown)


AmeerJ97:
A few extra verification details:

- Compared the table directly against `manimgl --help`.
- Confirmed that the removed `--save_pngs` entry is stale: `manimgl --save_pngs` exits with code 2 and reports `unrecognized arguments: --save_pngs`.
- Restored the original grid-table formatting and surrounding copy; the final commit is limited to 9 additions and 5 deletions in the flag table.
- `git diff --check` passes. A local Sphinx run currently stops before parsing the page because `sphinx_copybutton` is not installed in this checkout.

Let me know if anything needs to be adjusted!

00PrabalK00:
LGTM from a read-through. I checked each added/removed flag against `manimlib/config.py`:

- `--save_pngs` (`-g`) and `--config` are genuinely gone from the parser, so the removals are right.
- All the additions exist: `--vcodec`, `--pix_fmt`, `--subdivide`, `--show_animation_progress`, `--prerun`, `--clear-cache`.
- `--pix_fmt`'s documented `yuv420p` default matches `pixel_format` in `manimlib/default_config.yml`.
- The `--embed` rewrite is a real correction rather than a rewording — the old text described the removed string-insertion behavior, while the flag now takes a line number (`config.py` does `int(args.embed)`, and `extract_scene.py` inserts `self.embed()` at that line).

The table also still lines up: columns start at the same offsets and all three border rows are the same length. Parsing the file with docutils standalone gives no table errors, only the expected `Unknown interpreted text role "doc"` from the Sphinx-only role — so the `sphinx_copybutton` failure you hit isn't hiding a markup problem.

Two optional nits, neither blocking:

1. `--start_at_animation_number START[,END]` reads better, but argparse's auto-generated metavar is actually `START_AT_ANIMATION_NUMBER`, so that one entry doesn't literally match `manimgl --help`.
2. Pre-existing drift you may or may not want in scope here: `-s` is documented as "Skip to the last frame" while the parser help says "Save the last frame", and `--quiet` still has an empty function cell because its `help` is an empty string.

AmeerJ97:
Thanks for the careful pass. I incorporated the two drift items: `-s` now matches the parser wording, and `--quiet` has a behavior-based description. I kept `START[,END]` intentionally as the clearer user-facing input shape rather than argparse's generated variable name, and adjusted the PR description so it no longer claims a literal transcription of `manimgl --help`.

I also rebased the branch onto the current `master` and reran the option-set, grid-table, docutils, and whitespace checks.

docs/source/getting_started/configuration.rst
```diff
@@ -38,31 +38,35 @@ flag                                                       abbr   function
 ``--help``                                                 ``-h`` Show the help message and exit
 ``--version``                                              ``-v`` Display the version of manimgl
 ``--write_file``                                           ``-w`` Render the scene as a movie file
-``--skip_animations``                                      ``-s`` Skip to the last frame
+``--skip_animations``                                      ``-s`` Save the last frame
```
3b1b:
```suggestion
``--skip_animations``                                      ``-s`` Skip to the last frame
```

3b1b: The -s flag does not save a frame, that only happens using -w or -o

PR #2522, opened 2026-09-02
Handle video device source in get_full_video_path

MathItYT:
## Motivation
This pull request introduces a small but important change to the `get_full_video_path` function in `manimlib/utils/images.py`. The function now checks if the provided `video_file_name` is a digit (indicating a video device source, like a camera) and returns it directly instead of searching for a file path.

* [`manimlib/utils/images.py`](diffhunk://#diff-d32476c68788f7248923e9e8a7adcb6b4ba58bf48cd3f29859f428acf5a5de63R26-R27): Updated `get_full_video_path` to immediately return device source identifiers (when the input is all digits) rather than attempting to locate a file.

## Proposed changes
- `manimlib/utils/images.py`

### DIFF
--- a/manimlib/mobject/types/video_mobject.py
+++ b/manimlib/mobject/types/video_mobject.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from collections import OrderedDict
 from fractions import Fraction
+import sys
 
 import av
 import numpy as np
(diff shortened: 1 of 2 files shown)


MathItYT:
Wait, it doesn't work in every OS, only for MacOS and I should enable live streaming rather than preloading as the camera is a live stream, not a pre-recorded video.

MathItYT:
It's ready, but not compatible with Windows yet

